PR #17, opened 2017-11-25
Update workspace filenames

dabbott:
Dropping the "cs" prefix and renamed to Lona


dabbott:
cc @ryngonzalez you'll need to rename your workspace files

PR #21, opened 2017-11-26
Use new workspace filenames

haswalt:
Updates the generator code to use the new workspace filenames used by the studio


dabbott:
Thanks! Btw, I'm planning on rewriting the main component generation part of the react generator this week. It's pretty hacky and out of date with the way logic works now.

PR #23, opened 2017-11-27
[DOCS] Updated README.md to fix twitter link to '@dvnabbott'

lennartstolz:
In 9111af5b7b85333401c2daf174d55ef497272410 the twitter username of '@dvnabbott' was updated.  However the link still targeted the previous username '@devinaabbott'.

That's fixed in this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,6 +170,6 @@ Time constraints 😅
 
 ## The Team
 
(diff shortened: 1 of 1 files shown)


dabbott:
Thanks :D would never have caught this

PR #31, opened 2017-12-06
Text style picker

NghiaTranUIT:
## Changed Logs
+ Implement `NSViewHoverExtensions.swift` to support Hover Behavior
+ Update to macOS 10.12 to get rid of #available if (need discuss, since we're in dogfooding process, we could ship to mac 10.11 if we have enough human resource)
+ Override fontSize (but keep another text attribute) to make the title fits properly in `TextStylePickerButton`
+ Resolve cycle retain in some closures
+ Support dynamic size for NSPopover (depend on length of AttributeText inside)
+ Support dynamic size for `TextStyleItemView`

## Screenshots
screen shot 2017-12-06 at 9 55 05 pm

Looking forward to hearing your feedback 👍


LonaStudio/Utils/YogaLayout.swift
```diff
@@ -125,6 +125,11 @@ extension YGNodeRef {
         set { return YGNodeStyleSetAlignItems(self, newValue) }
     }
     
+    var alginSelf: YGAlign {
```
dabbott: typo: should be `alignSelf`

LonaStudio/Utils/AttributedFont.swift
```diff
@@ -39,14 +39,22 @@ public class AttributedFont {
         self.lineBreakMode = lineBreakMode
     }
     
-    public var nsFont: NSFont {
-        if let targetFont = NSFontManager.shared().font(withFamily: fontFamily, traits: NSFontTraitMask(rawValue: 0), weight: weight.rawValue, size: fontSize) {
-            return targetFont
+    public func nsFont(_ fontSize: CGFloat) -> NSFont {
```
dabbott: Since `fontSize` is a member variable of `AttributedFont`, I don't think we should allow overriding it here. The `AttributedFont` instance should be updated before calling `nsFont`.

E.g. in the caller:

```swift
let attributedFont = /* some font */

attributedFont.fontSize = 14

let whatever = attributedFont.nsFont
```

LonaStudio/Utils/AttributedFont.swift
```diff
@@ -79,13 +87,12 @@ public class AttributedFont {
         return paragraphStyle
     }
     
-    private func attributeDictionary() -> [String: Any] {
+    private func attributeDictionary(_ size: CGFloat? = nil) -> [String: Any] {
```
dabbott: Same feedback as my comment above on `nsFont`

NghiaTranUIT:
Thank for pointing this issues.
I will take a look how we could reuse ColorPickerView in TextStylePickerView 👍

NghiaTranUIT:
Personally, I'd merely copy/pass and rename the `ColorList` and `ColorSwatchView` with slight modification in order to implement the Text Style Picker.
Then I will create new PR to implement new General Picker View then refactoring `ColorPicker` and `TextStylePicker`. From that, we can reuse GeneralPickerView for any purpose ;]
How do you think ?

dabbott:
@NghiaTranUIT yes! That sounds like a perfect approach 😃

NghiaTranUIT:
@dabbott I've just updated and improve your last feedback.

**Changed log**
+ Support Search functionality
+ Same behaviour like ColorPickerView
+ Revert change in AttributeFont and YGNode
+ Address your feedback

Looking to hearing your thought 👍

Issue #33, opened 2017-12-06
Canvas heights not set by default


dennislaupman:
Hi,

When I open a new project, The canvas width is set by default to the correct 375px, but the height doesn't. The height is set to 100px. Is there a reason for that?

![image](https://user-images.githubusercontent.com/1274858/33685953-8423c21c-dad3-11e7-9765-4b888ecba283.png)

Otherwise I will change it and send a pull request :P

Grtz Dennis

dennislaupman:
See pull request #34

dennislaupman:
Hello Devin,

>> The main use case of Lona Studio (for us right now, at least) is designing individual components,

Ahhhh that clarifies a lot.

I also have some ideas about this. I will work this out and post it.

What I wondered. Where are the parameter values set? The demo texts, images etc? I can see them in the demo component (card). But not in the GUI.

Greetings Dennis

PS. Is it an idea to setup a Gitter?  Thats works great for brainstorms

dabbott:
Let me think about the Gitter a bit... I'm thinking _not yet_ since I'm not sure I'll have the time to stay on top of it at the moment. But this could be a good idea when I have a little more bandwidth in the future. Let's stick to Github issues for now but revisit if it becomes a hassle.

The parameters come from *Cases* in this example (definitely need a better way to visualize where parameters come from)

![screen shot 2017-12-07 at 2 11 03 pm](https://user-images.githubusercontent.com/1198882/33741282-8f3f53c0-db58-11e7-96d1-ad6b802d0394.png)

haswalt:
So far I see 2 good solutions:

1. Given that a  lot of the UI decision have been taken from Sketch start with no canvases defined. this implies that the canvas is based on your decisions.
2. Start with (n) clearly labeled phone sizes. This will give a guide to new users to what they may be building, coming from design land I often start with a full screen design and break down into components once I have an idea of the styling. Most small agencies and individuals don't have a full DLS setup before starting a project.

dennislaupman:
Hi Devin, Haswalt,

I think it is easier to give them the choice of canvas or component for they begin. See my (quick) concept.

![welcome_to_lona](https://user-images.githubusercontent.com/1274858/33744239-9c333080-dbb0-11e7-9423-a57902d4fcaf.png)

After they make a choice we can set te canvas to the immediately height.

dabbott:
Right, creating two different starting "templates" (Canvas/Screen or Component) seems reasonable.

#### Pros
[...]

PR #34, opened 2017-12-07
Set the correct default heights

dennislaupman:
When you open the Lona application the first time it is very confusing that the "Artboards" has not the right height. With the correct default heights it is immediately clear that these represent the different (iphone) devices.


haswalt:
Considering the discussion in #33 I think that this is a good interim step to make things clearer until a decision can be made.

dabbott:
My main use case is the "component" case for the Airbnb design system, so this wouldn't be the right default for me 😢 . Let's see if we can figure out something better in #33.

PR #40, opened 2017-12-09
Support Shadow Picker

NghiaTranUIT:
I branded off from your `shadows` branch and continue my work
Thank for your shadow model logic 🤣

**Change logs**
+ Support ShadowList and ShadowPicker (reused from ColorPickerView)
+ Reflect shadow attribute 
+ Only support for UIView

**Screenshots**
screen shot 2017-12-09 at 3 29 01 pm
screen shot 2017-12-09 at 3 28 54 pm

Looking forward to hearing your thought 👍

### DIFF
--- a/LonaStudio/Workspace/InspectorView.swift
+++ b/LonaStudio/Workspace/InspectorView.swift
@@ -53,6 +53,10 @@ class InspectorView: NSStackView {
         case backgroundColorEnabled
         case backgroundGradient
         
+        // Shadow
+        case shadowEnabled
+        case shadow
+        
         // Text
         case text
         case textStyle
@@ -136,6 +140,9 @@ class InspectorView: NSStackView {
     var borderColorButton = ColorPickerButton(frame: NSRect.zero)
     var borderColorEnabledView = CheckboxField(frame: NSRect.zero)
     var borderRadiusView = NumberField(frame: NSRect.zero)
+    var shadowEnabledView = CheckboxField(frame: NSRect.zero)
+    var shadowButton = ColorPickerButton(frame: NSRect.zero)
+    
     var textStyleView = TextStylePickerButton(frame: NSRect.zero)
     var numberOfLinesView = NumberField(frame: NSRect.zero)
     
@@ -501,6 +508,22 @@ class InspectorView: NSStackView {
         return backgroundSection
     }
     
+    func renderShadowSection() -> DisclosureContentRow {
+        shadowEnabledView.imagePosition = .imageOnly
+        
+        let shadowContainer = NSStackView(
+            views: [
+                shadowEnabledView,
+                shadowButton,
+                ],
+            orientation: .horizontal,
+            stretched: true
+        )
+        
+        let backgroundSection = renderSection(title: "Shadow", views: [shadowContainer])
+        return backgroundSection
+    }
+    
     func renderTextSection() -> DisclosureContentRow {
         textView.usesSingleLineMode = false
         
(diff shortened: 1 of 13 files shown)


LonaStudio/Workspace/InspectorView.swift
```diff
@@ -501,6 +508,22 @@ class InspectorView: NSStackView {
         return backgroundSection
     }
     
+    func renderShadowSection() -> DisclosureContentRow {
+        shadowEnabledView.imagePosition = .imageOnly
+        
+        let shadowContainer = NSStackView(
```
dabbott: Since `View`s don't support shadows yet, we should hide this like we do for a few of the other properties: https://github.com/airbnb/Lona/blob/master/LonaStudio/Workspace/InspectorView.swift#L665

NghiaTranUIT:
All of feedbacks are correct, I will update it soon.
About the clipped shadow, I suppose it' super easy to make the shadow doesn't clip by view. Just set `masksToBounds` on CALayer, and it's done ;]

NghiaTranUIT:
Hm, actually, it's hard than I thought. For some reasons, `textLayer.layer?.masksToBounds = false` & `wantLayer = true` result in disappearing NSAttributeString 🤔 
Maybe I should override CSTextView and draw shadow manually

NghiaTranUIT:
After investigating carefully, I turn out the problem is the CSTextView clips the contents.

As you can see the screenshot, I only add 1 subview is TextView
screen shot 2017-12-12 at 9 51 36 pm

However, when I capture the layer, there are 2 views which represent for the textView I added.

Interestingly enough, only CSView's heigh expands when I set fixed height. On the other hand, the CSTextView doesn't expand, it's still small size.
As a result, it clips the shadow.

-------
I tried to hard code the height when rendering the TextView. But it still doesn't work ?
https://github.com/airbnb/Lona/blob/957e893236d5fab84e89d11f7c3754f86b1e6c1f/LonaStudio/Canvas/CanvasView.swift#L106-L111

@dabbott Do you have any idea why CSTextView wasn't expand the height 🤔

dabbott:
@NghiaTranUIT I'm not sure... I'll checkout your branch and take a look now.

PS: I added you to the README. Feel free to change what I put.

dabbott:
I might have to play around with this a bit. Either way, when I'm done I'll merge it :+1:. It looks good and we can always iterate.

NghiaTranUIT:
Yes, Thank you in advance @dabbott, I spent to the shadow issue for a while, but no clue. Let me know if you'd want a hand 👍

dabbott:
@NghiaTranUIT definitely a tricky problem. I'm going to remove the parent `CSView` from the view hierarchy, and expand the frame of the text. E.g.

```swift
let textLayer = renderTextLayer(layer: layer, width: width, height: height)

if let shadow = getLayoutShadow(layout: layer) {
    let padding = NSSize(
        width: abs(shadow.x) + shadow.blur,
        height: abs(shadow.y) + shadow.blur
    )
    textLayer.textContainerInset = padding
    textLayer.frame.origin.x -= padding.width
    textLayer.frame.origin.y -= padding.height
    
    // Conceptually it seems like we should multiply by 2,
    // but that doesn't add enough padding.
    textLayer.frame.size.width += padding.width * 4
    textLayer.frame.size.height += padding.height * 4
}
```

This gives the desired behavior:

screen shot 2017-12-20 at 3 18 37 pm

It's not an ideal solution, but I think it should be fine. I've been meaning to make a few other tweaks to Text (Text shouldn't be able to have padding/background/border), so I'm going to merge what you have and I'll do all this separately. I'll make a PR and tag you when I'm done.

NghiaTranUIT:
That's cool @dabbott. Thank for helping me 👍

PR #41, opened 2017-12-11
chore(ruby): Add ruby version file for consistent pod installs

ryngonzalez:
I was encountering this issue (https://github.com/CocoaPods/CocoaPods/issues/4771) when running `bundle exec pod install`: the Sparkle dependency was having quotes added around it due to a difference in Ruby version I was using and that was used to generate the `Podfile.lock`.

Setting a consistent version will prevent that issue from happening.


dabbott:
Great. Honestly I don't know enough about ruby to know what exactly this does -- I'm not using rvm, if that's what reads this. But definitely seems like the right move.

ryngonzalez:
Ah I should’ve added more context. This is used by rbenv (https://github.com/rbenv/rbenv) which seems to be the current popular way to manage your ruby versions. 

Will make a note to update read me so folks know that if you have rbenv installed, this will be read by it.

ryh:
@ryngonzalez 
rbenv install 2.0.0-p648
>WARNING: ruby-2.0.0-p648 is past its end of life and is now unsupported.
It no longer receives bug fixes or critical security updates.

will you upgrade ruby version and `.ruby-version`

Issue #36, opened 2017-12-07
"SyntaxError: Unexpected token )" when running compiler


lyahdav:
Following the [README](https://github.com/airbnb/Lona#generating-code), I ran:
```
cd compiler/react
node index.js [my_workspace_path] output
```

And I got the following error:
```
compiler/react/index.js:21
  )
  ^
SyntaxError: Unexpected token )
    at Object.exports.runInThisContext (vm.js:78:16)
    at Module._compile (module.js:543:28)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)
    at Function.Module._load (module.js:439:3)
    at Module.runMain (module.js:605:10)
    at run (bootstrap_node.js:420:7)
    at startup (bootstrap_node.js:139:9)
    at bootstrap_node.js:535:3
```

I'm on node 7.4.0.

lyahdav:
Resolved by using node v8.9.2.

PR #42, opened 2017-12-11
chore(javascript): Add nvmrc to project so that we can have consistent JS use

ryngonzalez:
Prevent issues like #36 from happening.


.nvmrc
```diff
@@ -0,0 +1 @@
+8.9.2
```
ljharb: this can probably just be `8`.

Also, the file requires a trailing newline (like most unix files), or else `nvm` won't work properly.

ryngonzalez: Updated, @ljharb!

dabbott:
Good thinking!

PR #49, opened 2017-12-21
Update README.md

dennislaupman:
## What

Updated documentation

## Why

Sketch would work without this installation

## Major Changes

* I added the modules for this change…
* And this other thing too!

## Requirements for Merging

- [ ] It should do this
- [ ] It should edit this file
- [ ] It should look like this

## Testing Plan

- [ ] Checked that existing features work

## Feedback

I'm looking for feedback on:

* <this important decision>
* <this design>
* <this code style>

cc: @dabbott, @ryngonzalez

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,15 @@ Open LonaStudio.xcworkspace and build in Xcode 9 on Sierra+. If there are warnin
 
 > It will build on El Capitan, but it likely won't be usable. The changes needed to make are small, if anybody wants to add support.
 
+### Component-to-Sketch
+
+To use the component export to Sketch functionality. You first have to install some node modules. 
+
+```
+$ cd /Lona/LonaStudio/Scripting/Modules/component-to-sketch
+$ npm install
+```
+
 ## Workspace
 
 To work in Lona Studio, you'll need a properly configured *workspace*. A workspace is a directory containing, optionally, the following files:


dabbott:
Honestly I haven't tested this in a while... does the sketch export still work? If it works, we can document it. Otherwise, I probably wouldn't yet.

I think the right solution here is to commit the `node_modules`. Lona Studio should be able to run this script via the GUI, so the `node_modules` will need to be installed already:

screen shot 2017-12-21 at 11 48 36 am

dennislaupman:
Yes after the NPM it works nice!

dennislaupman:
I will commit and create a pull request

dennislaupman:
Hmmmzzz I thing commiting  the node module folder isn't an option ;) 

image

ljharb:
Committing node_modules is never a good idea.

CI and users (or app code) should run `npm install`. If you need determinism, use a lockfile.

dennislaupman:
@ljharb: True ;) That was also my opinion ;-) I'm checking out right now what is the best moment to run the npm -i

dabbott:
I personally don't have any problem with committing node_modules as a super low effort solution, but agree that installing them as part of a hook is a better plan.

Just remember @dennislaupman that we either need to build the node_modules against node 8.4 (the one bundled in Lona) or remove the native dependencies (node-obj-c) entirely.

dabbott:
@dennislaupman I'm going to close this, since it's been open for a while, and the task changed a bit. We should still figure out a good way to make the export-to-sketch command work by automatically installing npm modules at some point -- but we can open a separate issue for it.

Issue #63, opened 2018-01-18
Support TravisCI and Danger


NghiaTranUIT:
## Suggesting a New Feature

In order to keep this repository consistently, and enforce how the people contribute to the project.
I'd like to introduce the Danger in conjunction with TravisCI.

## Summary of Idea

**TravisCI**
One of famous CI server. It's totally free for OSS like Lona. We leverage the CI system in order to run test and build automatically with effortless.

**Danger**
Human can't possibly check everything. For instance, the contributor forget to 
+ Write unit-test
+ Code convention warning by SwiftLint
+ Warning if it's big PR
+ Warning for WIP for Don't Not Merge PR
+ ...

So let the Danger bot do it. It's potential candidate when conjunction with TravisCI.

### Travis
We can inspire how Alamofire construct the [.travis.yml](https://github.com/Alamofire/Alamofire/blob/master/.travis.yml) file.
It should be
```
language: objective-c
osx_image: xcode9
rvm:
  - 2.2.2

env:
  global:
    - WORKSPACE="LonaStudio.xcworkspace"
    - OSX_APP_SCHEME="LonaStudio"
    - OSX_SDK=macosx10.13
  matrix:
    - DESTINATION="arch=x86_64"                      SCHEME="$OSX_APP_SCHEME"           SDK="$OSX_SDK"     RUN_TESTS="YES"    BUILD_EXAMPLE="YES"

before_install:
  - bundle install
  - bundle exec pod repo update
  - bundle exec pod install

script:
  - set -o pipefail
  - xcodebuild -version
  - xcodebuild -showsdks
  - xcodebuild -list

    # Build Lona
  - if [ $BUILD_EXAMPLE == "YES" ]; then
      xcodebuild -workspace "$WORKSPACE" -scheme "$SCHEME" -destination "$DESTINATION" -configuration Debug ONLY_ACTIVE_ARCH=NO CODE_SIGN_IDENTITY="" CODE_SIGNING_REQUIRED=NO build | xcpretty;
    fi

    # Run Tests
  - if [ $RUN_TESTS == "YES" ]; then
      xcodebuild -workspace "$WORKSPACE" -sdk "$SDK" -scheme "$SCHEME" -destination "$DESTINATION" -configuration Debug ONLY_ACTIVE_ARCH=NO ENABLE_TESTABILITY=YES CODE_SIGN_IDENTITY="" CODE_SIGNING_REQUIRED=NO test | xcpretty;
    fi

after_success:
[...]

PR #67, opened 2018-01-20
Travis Danger Support

NghiaTranUIT:
## What

Refer from #63 

## Major Changes

* Update Gemfile, Podfile
* Add Dangerfile and .travis.yml

cc: @dabbott, @ryngonzalez


Dangerfile
```diff
@@ -0,0 +1,29 @@
...
+end
+
+# Warn when library files has been updated but not tests.
+tests_updated = !git.modified_files.grep(/Tests/).empty?
+if has_app_changes && !tests_updated
+  warn("The library files were changed, but the tests remained unmodified. Consider updating or adding to the tests to match the library changes.")
+end
+
+# Run SwiftLint
+#swiftlint.lint_files inline_mode: true
```
NghiaTranUIT: Intents for executing the SwiftLint in the future.

NghiaTranUIT:
screen shot 2018-01-20 at 3 51 17 pm

It seems that the Travis is hanging on forever 😂

I tried to run on my forks, had same problem. I suppose it's from Travis side.

Can you @dabbott double-check the Travis's Setting or somethings ?

NghiaTranUIT:
It's finally pass the Travis 😍

.travis.yml
```diff
@@ -0,0 +1,37 @@
...
+rvm:
+  - 2.2.2
+
+env:
+  global:
+    - WORKSPACE="LonaStudio.xcworkspace"
+    - OSX_APP_SCHEME="LonaStudio"
+    - OSX_SDK=macosx10.13
+  matrix:
+    - DESTINATION="arch=x86_64"                      SCHEME="$OSX_APP_SCHEME"           SDK="$OSX_SDK"     RUN_TESTS="YES"    BUILD_EXAMPLE="YES"
```
ljharb: rather than running tests and building the example in the same job, this should be split into multiple jobs - either by using the matrix (if it's important to build the example on every ruby version) or by using `include:` (ie, explicitly included builds)

NghiaTranUIT: You're right. I've just updated the matrix in order to support parallel jobs 👍

.travis.yml
```diff
@@ -0,0 +1,41 @@
...
+  - 2.2.2
+
+env:
+  global:
+    - WORKSPACE="LonaStudio.xcworkspace"
+    - OSX_APP_SCHEME="LonaStudio"
+    - OSX_SDK=macosx10.13
+  matrix:
+    - DESTINATION="arch=x86_64"                      SCHEME="$OSX_APP_SCHEME"           SDK="$OSX_SDK"     RUN_TESTS="NO"    BUILD_EXAMPLE="YES"
+    - DESTINATION="arch=x86_64"                      SCHEME="$OSX_APP_SCHEME"           SDK="$OSX_SDK"     RUN_TESTS="YES"    BUILD_EXAMPLE="NO"
```
ljharb: you can make this a bit shorter by omitting the "no" entirely; and you may want to put "RUN_TESTS" and "BUILD_EXAMPLE" first, so it's clearer in the travis UI what's happening.

PR #72, opened 2018-01-23
refactor-swiftlint-rule-trailing-whitespace

huyphams:
## What

- I uncommented the first rule of SwiftLint and fixed thousands of warnings by this rule.

## Why

- This is one of a common rule and it's easy to fix.

## Major Changes

- Remove space
- Remove `,`
- Update .gitignore added ignore OS file: .DS_Store

cc: @dabbott, @ryngonzalez, @NghiaTranUIT


dabbott:
Great! Thanks for doing this.

I think we should remove the changes from the `examples/generated/test/swift` directory, since this code is generated by the Lona compiler (and whitespace changes are meaningful).

Also, we might want to move the changes to the .pbxproj and Podfile.lock into a different commit. But I'm not that concerned for this change, so either way is fine.

huyphams:
Yup, I will update it.

huyphams:
Hi @dabbott I have updated the commit, can you review it 😄

PR #75, opened 2018-01-26
Clean SwiftLint warnings

NghiaTranUIT:
## Why

There are tons of annoying warnings. It drives me nut 😂

## Major Changes

* Disable `todo` warnings
* Resolve all hundreds of violent warnings.
* Set back Swift 3 @objc inference to `default` value (default behavior from Swift 4)

## Testing Plan

- [x] Tested this change locally
- [x] Checked that existing features work

cc: @dabbott, @ryngonzalez


LonaStudio/Models/CSData.swift
```diff
@@ -229,7 +223,7 @@ enum CSData: Equatable, CustomDebugStringConvertible {
         case .Null:
             return NSNull()
         case .Bool(let value):
-            return NSNumber(booleanLiteral: value)
+            return NSNumber(value: value)
```
dabbott: I'm pretty sure this does actually need to be `booleanLiteral` or it won't turn into a bool when written as JSON.

examples/generated/test/swift/style/TextStyles.swift
```diff
@@ -22,7 +22,6 @@ public class TextStyles: UIView {
```
dabbott: Files in `examples/generated`  are generated by the Lona compiler, so we shouldn't lint them. Can we add `examples` to the list of directories `excluded` by SwiftLint?

examples/generated/test/swift/logic/If.swift
```diff
@@ -1,50 +0,0 @@
-import UIKit
```
dabbott: Don't delete this file 😆

compiler/core/static/swift/AttributedFont.swift
```diff
@@ -17,8 +17,7 @@ public class AttributedFont {
     size: CGFloat = UIFont.systemFontSize,
```
dabbott: I think the changes in this file (and `examples/generated/test/swift/AttributedFont.swift`) are fine. This file is essentially a source file, since it's copied over into the generated code directory automatically. So I think we _should_ continue to lint the `compiler` directory for now. :+1:

dabbott:
Great job! Thanks for fixing these 😭 . The code is looking much more consistent already!

examples/generated/test/swift/logic/If.swift
NghiaTranUIT: Yeah, I just undo it now 😄

examples/generated/test/swift/style/TextStyles.swift
NghiaTranUIT: Yes sure 🔥

NghiaTranUIT:
I've roll-back all changed in `examples` and `compiler` as well as adding in `excluded` from SwiftLint 🍔

Issue #56, opened 2018-01-13
Undo/Redo Mechanism


NghiaTranUIT:
## Suggesting a New Feature

The undo/redo mechanism is a must-have functionality on any kind of design tool. So it's high priority to implement this feature carefully.

## Initial Design

In order to implement the Undo/Redo mechanism, we take an advantage of Command Design Pattern, which is one of famous pattern from Gang Of Four.

### Command
`Command` is abstract protocol, which defines the receiver and how to execute and undo an action

<img width="254" alt="screen shot 2018-01-13 at 9 55 20 am" src="https://user-images.githubusercontent.com/5878421/34902139-e385d7cc-f847-11e7-99ab-5d736134deb8.png">

All of change actions logic, such as UpdateValueCommand, UpdateColorCommand, UpdateShadowCommand, ... should be encapsulated into the `Command`'s adopting class.

<img width="933" alt="screen shot 2018-01-13 at 9 42 33 am" src="https://user-images.githubusercontent.com/5878421/34902048-2dd19afc-f846-11e7-9895-4bc5d46ba4a8.png">

### Global Command Queue
In order to support Undo/Redo mechanism globally, we define the queue, which holds 20 commands at most (configurable). It simply appends new command and pop out the latest command if it exceeds the threshold.
<img width="997" alt="screen shot 2018-01-13 at 9 43 31 am" src="https://user-images.githubusercontent.com/5878421/34902052-3c957450-f846-11e7-9d0f-2a3e25aa4729.png">

### Undo/Redo action
<img width="981" alt="screen shot 2018-01-13 at 9 43 42 am" src="https://user-images.githubusercontent.com/5878421/34902053-45163fec-f846-11e7-81cd-f28759654e77.png">
We define currentIndex, which is responsible for current action has been taken place. 

As a result, the `Undo` action simply invokes the `redo` action of current action, then moving the pointer back.
`Redo` action simply invokes the `execute` action of next command in the queue.

### New action while in middle of actions
<img width="981" alt="screen shot 2018-01-13 at 9 47 55 am" src="https://user-images.githubusercontent.com/5878421/34902101-2245a808-f847-11e7-91b5-09c874b478b3.png">
If the current pointer is in middle of the queue. It means the user does couple of redo, then executing new command. 

I suggest to drop all ahead actions and replace with new one. It's same behavior with Sketch or Photoshop.

## Proposed Implementation in Swift
`Command` protocol defines essential actions.
```swift
protocol Command {
[...]

PR #99, opened 2018-02-08
Undo Redo Mechanism 

NghiaTranUIT:
## Why

Refer to #56 

## Major Changes

* Introduce a lightweight UndoGlobalManager, which is the high level of NSUndoManager.
* Support undo logic in CSLayer,  LayerList
* Observe Undo/Redo notification and do reload appropriately.

## Testing Plan

- [x] Able to undo/redo in all variable of CSLayer
- [x] Able to undo/redo after adding/removing components in LayerList
- [x] Able to undo/redo after hiding/showing components
- [x] Global Undo/redo for each Document.
- [x] Undo/redo arrange action 
- [x] Undo/redo canvas properties
- [ ] Other stuffs.

## Feedback

I'm looking for feedback on the implementation and how I did Undo/redo on many properties 😄 

cc: @dabbott, @ryngonzalez

### DIFF
--- a/LonaStudio/CanvasList/CanvasListEditor.swift
+++ b/LonaStudio/CanvasList/CanvasListEditor.swift
@@ -111,12 +111,20 @@ class CanvasListEditor: NSOutlineView, NSOutlineViewDataSource, NSOutlineViewDel
         let canvas = item as! Canvas
         let cellView = NSTableCellView()
 
+        // Old value
+        let oldCanvas = canvas.copy() as! Canvas
+        let index = self.canvasList.index(where: { $0 === canvas })!
+        let undoFunc = {
+            self.canvasList[index] = oldCanvas
+            self.onChange(self.canvasList)
+        }
         switch tableColumn!.identifier.rawValue {
         case "Visible":
             let field = CSValueField(value: CSValue(type: CSType.bool, data: CSData.Bool(canvas.visible)))
             field.onChangeData = { value in
-                canvas.visible = value.boolValue
-                self.onChange(self.canvasList)
+                self.changedData(execute: {
+                    canvas.visible = value.boolValue
+                }, undo: undoFunc)
             }
             return field.view
         case "Name":
@@ -126,8 +134,9 @@ class CanvasListEditor: NSOutlineView, NSOutlineViewDataSource, NSOutlineViewDel
             field.value = canvas.name
 
             field.onChange = { value in
-                canvas.name = value
-                self.onChange(self.canvasList)
+                self.changedData(execute: {
+                    canvas.name = value
+                }, undo: undoFunc)
             }
 
             cellView.textField = field
@@ -142,8 +151,9 @@ class CanvasListEditor: NSOutlineView, NSOutlineViewDataSource, NSOutlineViewDel
             cellView.addSubview(field)
 
             field.onChange = { value in
-                canvas.width = value
-                self.onChange(self.canvasList)
+                self.changedData(execute: {
+                    canvas.width = value
+                }, undo: undoFunc)
             }
 
             return cellView
@@ -158,8 +168,9 @@ class CanvasListEditor: NSOutlineView, NSOutlineViewDataSource, NSOutlineViewDel
             cellView.addSubview(field)
 
             field.onChange = { value in
-                canvas.height = value
-                self.onChange(self.canvasList)
+                self.changedData(execute: {
+                    canvas.height = value
+                }, undo: undoFunc)
             }
 
             return cellView
@@ -168,18 +179,19 @@ class CanvasListEditor: NSOutlineView, NSOutlineViewDataSource, NSOutlineViewDel
             field.isBordered = false
 
             field.onChange = { value in
-                canvas.heightMode = value
-                self.onChange(self.canvasList)
+                self.changedData(execute: {
+                    canvas.heightMode = value
+                }, undo: undoFunc)
             }
 
             return field
         case "Scale":
             let field = PopupField(frame: NSRect.zero, values: ["1x", "2x"], initialValue: String(format: "%.0fx", canvas.exportScale))
             field.isBordered = false
-
             field.onChange = { value in
-                canvas.exportScale = value == "2x" ? 2 : 1
-                self.onChange(self.canvasList)
+                self.changedData(execute: {
+                    canvas.exportScale = value == "2x" ? 2 : 1
+                }, undo: undoFunc)
             }
 
             return field
@@ -187,17 +199,20 @@ class CanvasListEditor: NSOutlineView, NSOutlineViewDataSource, NSOutlineViewDel
             let colorValue = CSValue(type: CSColorType, data: CSData.String(canvas.backgroundColor))
             let field = CSValueField(value: colorValue)
             field.onChangeData = { value in
-                canvas.backgroundColor = value.stringValue
-                self.onChange(self.canvasList)
+                self.changedData(execute: {
+                    canvas.backgroundColor = value.stringValue
+                }, undo: undoFunc)
             }
+
             return field.view
         case "Parameters":
             guard let component = component else { break }
 
             let field = CSValueField(value: CSValue(type: component.parametersType(), data: canvas.parameters))
             field.onChangeData = { value in
-                canvas.parameters = value
-                self.onChange(self.canvasList)
+                self.changedData(execute: {
+                    canvas.parameters = value
+                }, undo: undoFunc)
             }
             return field.view
         default:
@@ -205,6 +220,13 @@ class CanvasListEditor: NSOutlineView, NSOutlineViewDataSource, NSOutlineViewDel
         }
 
         return cellView
+    }
+
+    private func changedData(execute: @escaping () -> Void, undo: @escaping () -> Void) {
+        UndoManager.shared.run(name: "Changed", execute: {
+            execute()
+            self.onChange(self.canvasList)
+        }, undo: undo)
     }
 
     var selectedItem: Canvas? {
--- a/LonaStudio/Controls/NumberField.swift
+++ b/LonaStudio/Controls/NumberField.swift
@@ -25,7 +25,6 @@ class NumberField: NSTextField, NSTextFieldDelegate, NSControlTextEditingDelegat
         set {
             stringValue = String(newValue)
             onChange(value)
-            onChangeData(CSData.Number(value))
         }
     }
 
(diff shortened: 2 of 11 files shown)


dabbott:
Ok, this is a great first step! 😄 Now we can see how all the different pieces work. I have a lot of feedback this time around since undo/redo is something we'll have to maintain and support with each new feature.

#### 1. Control flow

I find the overall flow of the code a little hard to follow, and I think we should see if we can make it more straightforward. 

Currently, when I undo a command, `CSLayer` records a redo command implicitly, then a notification is handled by the `ViewController` which tells the inspector to re-render. I'd like the relationship between model updates/undo/redo/re-render to be more explicit.

Also
- Recording commands from `CSLayer` directly will only work when properties are modified individually, which probably won't always be true (we could use command grouping maybe, but that's another layer of complexity)
- There's a _lot_ of boilerplate code for monitoring each property

I think it would be clearer if we record every command and how to undo it in the same place. Here's roughly what I'm thinking:

```swift
struct Command {
    let name: String
    let execute: () -> Void
    let undo: () -> Void
}

final class UndoGlobalManager: UndoGlobalManagerType {

    // ...

    // MARK: - Public

    func run(_ command: Command) {
        func undo() {
            command.undo()
            self.manager.registerUndo(withTarget: self) { _ in execute() }
        }
        func execute() {
            command.execute()
            self.manager.registerUndo(withTarget: self) { _ in undo() }
            self.manager.setActionName(command.name)
        }
        execute()
    }

    func run(name: String, execute: @escaping () -> Void, undo: @escaping () -> Void) {
        run(Command(name: name, execute: execute, undo: undo))
    }
}
```

Then we can `run` commands whenever the user interacts with a control that modifies the `CSLayer` model. For example, we can handle almost all updates to `CSLayer` by updating `InspectorView` from:

https://github.com/airbnb/Lona/blob/fa62235499f62de2a348bc82e61118a8f2e19dad/LonaStudio/Workspace/InspectorView.swift#L771-L775

To:

```swift
        fields.forEach({ (control, property) in
            var control = control
            if let value = properties[property] { control.data = value }
            // Assign `onChangeData` after assigning `data`
            control.onChangeData = { data in
                let oldValue = self.value[property]

                UndoGlobalManager.default.run(name: property.rawValue, execute: {
                    self.handlePropertyChange(for: property, value: data)
                }, undo: {
                    self.handlePropertyChange(for: property, value: oldValue ?? properties[property]!)
                })
            }
        })
```

I think this will be more straightforward to maintain. Every command will have both an `execute` and `undo` right next to each other, so it'll be hard to forget to implement/update `undo`. This will also handle re-rendering the inspector and canvas view automatically, so we don't have to use `NotificationCenter`. This would work pretty much the same way for `addLayer` and `removeLayer` too.

#### 2. Maybe remove `UndoGlobalManagerType`

Is anything else going to conform to `UndoGlobalManagerType`? If not, I'd probably remove it, and just put the methods directly on `UndoGlobalManager`. Or even better...

#### 3. Maybe use extension of `UndoManager` instead of `UndoGlobalManager`

Since `UndoGlobalManager` is mostly providing some utility functions for `UndoManager`, I think it might be clearer to put them on the `UndoManager` class directly. E.g.

```swift
extension UndoManager {
    static var shared: UndoManager {
        return NSDocumentController.shared.currentDocument!.undoManager!
    }

    func run(_ command: Command) {
        func undo() {
            command.undo()
            self.registerUndo(withTarget: self) { _ in execute() }
        }
        func execute() {
            command.execute()
            self.registerUndo(withTarget: self) { _ in undo() }
            self.setActionName(command.name)
        }
        execute()
    }

    func run(name: String, execute: @escaping () -> Void, undo: @escaping () -> Void) {
        run(Command(name: name, undo: undo, execute: execute))
    }
}
```

That's a lot of ideas... but let me know what you think!

NghiaTranUIT:
Sorry for late replying. I've been in travel.

Thanks @dabbott for pointing this out. I didn't realize We could put the Undo logic inside the InspectorView.

I admin that my approach has a huge drawback by introducing too many boilerplate codes. I agree that your solution is more better and concise than the original 🎉 👍 

Let me follow your approach then discussing later.

NghiaTranUIT:
@dabbott Pls take a look my change - if it's good. I'm willing to do the rest part 👍 🎉 😄

LonaStudio/Controls/NumberField.swift
```diff
@@ -24,8 +24,6 @@ class NumberField: NSTextField, NSTextFieldDelegate, NSControlTextEditingDelegat
         get { return Double(stringValue) ?? 0.0 }
         set {
             stringValue = String(newValue)
-            onChange(value)
```
NghiaTranUIT: I removed this line due to mutex crash -> When redo the action, it implicitly called twice
```swift
control.onChangeData = { data in }
```
then resulting in a crash in `InspectorView`
screen shot 2018-02-14 at 3 11 43 pm
I encountered this while editing Dimension properties. Looking forward any idea 🤔 🇹🇭

dabbott: I think modifying `control.data` within the `onChangeData` handler causes an infinite loop. Changing the text in the UI calls `control.onChangeData`, which calls our undo/redo handler, and within that we're setting `control.data = data`, which again calls `onChangeData(value)`.

I think removing `onChange()` and `onChangeData()` like you did is the right solution :+1:. Before this change, we only ever modified the input field from the UI. Now that we do it programmatically, we can't assume that it's automatically up-to-date in the UI.

You'll probably also need to update `func control()` to call `onChange` and `onChangeData` in order for the arrow keys to work: https://github.com/airbnb/Lona/blob/6e503e174d250975cd2b6c3ef4882418d36491a7/LonaStudio/Controls/NumberField.swift#L75

LonaStudio/Workspace/ViewController.swift
```diff
@@ -904,9 +937,18 @@ class ViewController: NSViewController, NSOutlineViewDataSource, NSOutlineViewDe
...
                 checkbox.value = layer.visible
-                checkbox.onChange = { value in
-                    layer.visible = value
-                    self.render()
+                checkbox.onChange = {[unowned self] value in
+
+                    let oldValue = layer.visible
+                    UndoManager.shared.run(name: "Visible", execute: {[unowned self] in
+                        layer.visible = value
+                        checkbox.state = value ? .on : .off
```
dabbott: We should do `checkbox.value = value` instead of using `.on` and `.off`

LonaStudio/Workspace/ViewController.swift
```diff
@@ -590,20 +604,39 @@ class ViewController: NSViewController, NSOutlineViewDataSource, NSOutlineViewDe
...
+    private func updateChildren(children: [CSLayer], for parent: CSLayer) {
+
+        // Change
+        parent.children = children
+
+        // Refresh
+        refreshDocument(true)
+    }
+
+    private func refreshDocument(_ shouldClearInspector: Bool = false) {
```
dabbott: In general I'd leave the parameter label on a boolean rather than using `_`, since it's easier to understand when reading a function call. E.g. `refreshDocument(shouldClearInspector: true)` explains what the boolean does

dabbott:
I think this looks great! Keep up the good work 😄 

We're probably going to run into issues where the color checkboxes don't undo/redo correctly, since they automatically set the color to transparent when changed. This might not be worth fixing yet though, since I suspect it'll get pretty complicated.

At some point, I want to move toward using more immutable data structures, which would make this sort of thing easier. E.g. if CSLayer were immutable, we could roll back the entire layer and not have to worry about how the different properties interact. But that's for another time!

LonaStudio/Workspace/ViewController.swift
NghiaTranUIT: Unfortunately, It's impossible to assign `Bool` to `NSControl.StateValue` from `checkbox.state` 😅

NghiaTranUIT:
I updated few changed and support Undo/Redo for `Re-Arrange Action` and `Canvases List`. Glad to hear your thought :D @dabbott

LonaStudio/Controls/NumberField.swift
dabbott: We still need to update `func control()` in `NumberField` to call `onChange` and `onChangeData` in order for the arrow keys to work:

LonaStudio/Workspace/InspectorView.swift
```diff
@@ -770,8 +770,30 @@ class InspectorView: NSStackView {
...
-            control.onChangeData = { data in self.handlePropertyChange(for: property, value: data) }
+
+            // Set default value before setting onChangeData in order to avoid calling twice
             if let value = properties[property] { control.data = value }
+
+            // Setup onChangeData
+            control.onChangeData = { data in
+
+                let oldValue = self.value[property]
+                print("==== OnChangeData property = \(property) value = \(data)")
```
dabbott: Need to remove this comment

LonaStudio/Workspace/ViewController.swift
dabbott: I actually made a subclass to solve this 🤣, just use `value` instead of `state`. I do `checkbox.value = layer.visible` a few lines up

https://github.com/airbnb/Lona/blob/master/LonaStudio/Controls/CheckboxField.swift#L21

LonaStudio/CanvasList/CanvasListEditor.swift
```diff
@@ -207,6 +222,13 @@ class CanvasListEditor: NSOutlineView, NSOutlineViewDataSource, NSOutlineViewDel
         return cellView
     }
 
+    private func onChangedData(execute: @escaping () -> Void, undo: @escaping () -> Void) {
```
dabbott: can we rename this to `changeData`? I use `on` in front of names to indicate that they are event/change handlers, but this is more of a utility method for calling a handler.

LonaStudio/Workspace/ViewController.swift
NghiaTranUIT: The only thing I concern, that if I update `checkbox.value`, it triggers the `onChange`, which result in recursive loop.

screen shot 2018-02-24 at 1 37 48 pm

Do you have any suggestion?

NghiaTranUIT:
The nightmare is coming. I tried to rebase from master. There are many conflict which I couldn't resolve 😂
Let me see if I could squash this PR and try rebasing again.

NghiaTranUIT:
Done - rebase `master` successfully 🍻 😄

dabbott:
Great 🚢! Very exciting to have this in 😄

PR #103, opened 2018-02-19
Refactor View Controller

NghiaTranUIT:
## What

The main `ViewController` is going to grow dramatically, and hold many unnecessary responsibilities. For example, 
1. LayerList logic and LayerList DataSource.
2. Setup Inspector View for component or layer.

## Why

I did this change in order to reduce the complexity and massive controller. Make the world is easier.

## Major Changes

- Refactor `ViewController`
- Remove all `LayerList's Data Source` from `ViewController`.
- Moved this logic to LayerList class.
- Introduce the `ComponentInspectorView`, which is responsible for handling the Component layer logic.
- Introduce `InspectorContentView`.

## Testing Plan

- [x] Tested this change locally
- [x] Checked that existing features work

## Summary 
The `ViewController` is now elegant and transparent than before.
Looking to hear your thought 🎉 😄 

cc: @dabbott, @ryngonzalez


dabbott:
Excellent! ViewController was getting out of hand 😎 . I'll look at this PR and your other one tomorrow.

PR #109, opened 2018-03-08
Add Record type editor

dabbott:
## What

You can now create custom Record types as parameters (necessary for current Airbnb components). Other complex type editing to come.

A record is a type that contains key-value pairs. This is useful for grouping parameters, or directly interoperating with complex types in the target codebase.

In JSON, record types are akin to:
`{ "title": "String", "subtitle": "String?" }`

## Testing Plan


ryngonzalez:
What is a record? Could we update the description here to get more context?

dabbott:
@ryngonzalez are we gonna start doing code reviews? 😂

Issue #83, opened 2018-01-30
Support View borderWidth and borderColor when generating Swift code


dabbott:
## Summary

Let's update the Swift code generation to support `borderWidth` and `borderColor`.

## Initial Design

Lona Studio already allows editing these styles:

![screen shot 2018-01-29 at 5 00 24 pm](https://user-images.githubusercontent.com/1198882/35542652-ee4f9af0-0515-11e8-8911-c197bd8fd600.png)

## Proposed Implementation

I'm trying to support both iOS and OSX where possible, so:
- On iOS, I think we can use the `borderWidth` and `borderColor` properties of `CALayer`. We'll just need to convert our `UIColor`s to `CGColor`s.
- On OSX, I've been avoiding layer-backed/hosted views, so we should go with the `borderColor` and `borderWidth` of `NSBox` (currently Lona `View` maps to `NSBox`). We'll also need to set `borderType = .lineBorder` I think.

For now, assume that only Lona `View` can have a `borderWidth` and `borderColor`. Adding a border to a Lona `Text` or `Image` is undefined behavior (even though Lona Studio allows it... I may remove).

#### Compiler

In the Reason code, we'll start by defining the type and default value of these new styles:

https://github.com/airbnb/Lona/blob/master/compiler/core/src/core/decode.re#L5-L31
https://github.com/airbnb/Lona/blob/master/compiler/core/src/core/lonaValue.re#L19-L46

Conveniently the names are the same as the names defined in JSON, but we'll need to convert the color we use to `cgColor` on iOS. The general idea is that you'll overwrite the `BinaryExpression` AST node to specifically catch and rewrite this case. As an example, here's how I convert `text`-related things to use `attributedText` and `attributedStringValue`:

https://github.com/airbnb/Lona/blob/master/compiler/core/src/swift/swiftLogic.re#L156

It's not the prettiest, but it works for now. I think once I have a better sense of all the possible use cases, I'll introduce an abstraction that isn't based on string replacement.

## Test Plan

To test, create a new component in the `examples/test` workspace that has a border style, and re-generate the swift code for this workspace from the `compiler/core` directory with `node src/main.bs.js workspace swift ../../examples/test ../../examples/generated/test/swift`.

To test for OSX, I haven't started saving the snapshots (I definitely should), but just run `node src/main.bs.js workspace swift ../../examples/test ../../examples/generated/test/osx --framework=appkit`

[...]

PR #132, opened 2018-03-28
Support View borderWidth and borderColor when generating Swift code

jasonzurita:
cc: @dabbott 

## What

Github issue: https://github.com/airbnb/Lona/issues/83

(screenshots below)

## Major Changes

* Adds borderWidth and borderColor support for both iOS and MacOS.
* MacOS generated files will continue to use a `borderType` of `.noBorder` when `borderWidth` is not set in the Lona component file. If it is set, `borderType` will change to `.lineBorder`.**

** See _Feedback_ section for more about this.

## Requirements for Merging

* Generated Swift code should match the Lona component file. There is a test component file and the respective generated Swift code included in this PR.

## Testing Plan

- [x] Tested this change locally.
- [X] Checked that existing features work.

Note: there will be another PR with an iOS / MacOS app to view the generated Swift code including the UI code generated in this PR.

## Feedback

* I'm looking for feedback on the Reason implementation (this is my first venture into using Reason, so I am sure there could be improvements to my implementation).

* Also for MacOS, if a child view has a borderWidth, the current implementation will flag the parent view as having a `borderType` of `.lineBorder` also. If no borderWidth was specified in the parent view, there will be a visible border because the default border width is 1. Some guidance on addressing this would be great or maybe this is acceptable for this PR.

### iOS
screen shot 2018-03-28 at 7 58 39 am

### MacOS
screen shot 2018-03-28 at 7 58 01 am

### DIFF
--- a/compiler/core/src/swift/swiftComponent.re
+++ b/compiler/core/src/swift/swiftComponent.re
@@ -451,6 +451,21 @@ let generate =
       |> List.map(defineInitialLayerValues)
       |> List.concat;
     };
+    /* contains & isPropertyUsed should be made more general if needed in other places */
+    let contains = (~value: 'a, theList: list('a)) => {
+      let f = (found, elem) => found || elem == value;
+      ListLabels.fold_left(~f, ~init=false, theList);
+    };
+    let isPropertyUsed = (root, property) => {
+      let bindings = 
+        root
+        |> Layer.flatten
+        |> List.map((v: Types.layer) => v.parameters |> StringMap.bindings)
+        |> List.flatten
+        |> List.map(((k, _)) => k);
+
+        contains(~value=property, bindings);
+    };
     let resetViewStyling = (layer: Types.layer) =>
       switch layer.typeName {
       | View => [
@@ -463,7 +478,7 @@ let generate =
             "left":
               layerMemberExpression(layer, [SwiftIdentifier("borderType")]),
             "operator": "=",
-            "right": SwiftIdentifier(".noBorder")
+            "right": isPropertyUsed(layer, "borderWidth") ? SwiftIdentifier(".lineBorder") : SwiftIdentifier(".noBorder")
           }),
           BinaryExpression({
             "left":
--- a/compiler/core/src/swift/swiftComponent.re
+++ b/compiler/core/src/swift/swiftComponent.re
@@ -451,20 +451,16 @@ let generate =
       |> List.map(defineInitialLayerValues)
       |> List.concat;
     };
-    /* contains & isPropertyUsed should be made more general if needed in other places */
-    let contains = (~value: 'a, theList: list('a)) => {
-      let f = (found, elem) => found || elem == value;
-      ListLabels.fold_left(~f, ~init=false, theList);
-    };
-    let isPropertyUsed = (root, property) => {
-      let bindings = 
-        root
-        |> Layer.flatten
-        |> List.map((v: Types.layer) => v.parameters |> StringMap.bindings)
-        |> List.flatten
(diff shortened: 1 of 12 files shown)


jasonzurita:
Huge thanks to @dabbott for your help on this!

compiler/core/src/swift/swiftComponent.re
```diff
@@ -451,6 +451,21 @@ let generate =
...
       |> List.concat;
     };
+    /* contains & isPropertyUsed should be made more general if needed in other places */
+    let contains = (~value: 'a, theList: list('a)) => {
+      let f = (found, elem) => found || elem == value;
+      ListLabels.fold_left(~f, ~init=false, theList);
+    };
+    let isPropertyUsed = (root, property) => {
+      let bindings = 
+        root
```
dabbott: If I'm understanding the child/parent border width issue you're mentioning correctly, I think you need to make sure that `isPropertyUsed` is checking just the specific layer it's called with, rather than the entire layer hierarchy. You'll also want to detect if one of these parameters is assigned by logic.

I tried it out to double check and I think this should be what you want:

```reason
    let isPropertyUsed = (layer: Types.layer, property) => {
      let assignedParameters =
        Layer.LayerMap.find_opt(layer, layerParameterAssignments);
      let parameterIsAssigned =
        switch assignedParameters {
        | Some(parameters) => StringMap.mem(property, parameters)
        | None => false
        };
      parameterIsAssigned || StringMap.mem(property, layer.parameters);
    };
```

Side note: probably can use `List.find` or `List.exists` for `contains`.

jasonzurita: Nice, I will give this a try shortly!

jasonzurita:
@dabbott sorry for the delay in getting back to this. Your suggestions worked great!

screen shot 2018-04-04 at 6 19 12 pm

compiler/core/src/swift/swiftComponent.re
```diff
@@ -451,6 +451,17 @@ let generate =
...
+    /* isPropertyUsed should be made more general if needed in other places */
+    let isPropertyUsed = (layer: Types.layer, property) => {
+      let assignedParameters =
+        Layer.LayerMap.find_opt(layer, layerParameterAssignments);
+      let parameterIsAssigned =
+        switch assignedParameters {
+        | Some(parameters) => StringMap.mem(property, parameters)
+        | None => false
+        };
+      parameterIsAssigned || StringMap.mem(property, layer.parameters);
```
jasonzurita: @dabbott out of curiosity what does the `|| StringMap.mem(property, layer.parameters)` cover?

dabbott: Parameters are in two categories: some are set as the initial value for a layer, and some are assigned when executing logic (the `update` function). `StringMap.mem(property, layer.parameters)` covers the initial values

dabbott: @jasonzurita

dabbott:
@jasonzurita ok! I'm gonna merge this. 

I think there may be one more edge case to handle, which is when there's a border width that's conditionally assigned by logic, but not part of the layer's parameters. We'll probably need to default the border width to 0 in the `setUpViews` method. Let's just do that in another PR though, I think this is already solid for a first pass.

compiler/core/src/swift/swiftComponent.re
jasonzurita: ahhh, good to know, thanks!

jasonzurita:
@dabbott sounds good. Thank for your help with this PR!!

PR #133, opened 2018-04-09
Test iOS & MacOS apps to view generated Swift code

jasonzurita:
rr: @dabbott

## iOS & MacOS Apps for Testing

This Xcode project includes two targets: iOS & MacOS. Both may be used to view the generated Swift UI code to verify proper Lona compilation (layout, color, assignments, etc.).
- Changes to existing generated Swift code will automatically be reflected in the Xcode project!
- New Swift generated code files will need to be added as a reference to the Xcode project for testing.
- Both projects use a `Generated` enum that lists all the UI elements to be tested. If a new UI element is added to the `Generated` enums file, it will automatically be pulled into the respective table views for testing. iOS & MacOS have their own separate `Generated` enum files. (future work may be done to better combine these two, but this was considered to be acceptable for this first pass at the testing apps)

## Why

This is to enable easier testing of generated Swift code (visual & code via warning/errors).

## Major Changes

Add Xcode project for testing that includes iOS & MacOS apps.

## Testing Plan

## iOS App Screenshots
screen shot 2018-04-09 at 1 50 32 pm
screen shot 2018-04-09 at 1 50 42 pm

## MacOS App Screenshot
screen shot 2018-04-09 at 1 25 40 pm


examples/generated/test/appkit/images/LocalAsset.swift
```diff
@@ -46,7 +46,8 @@ public class LocalAsset: NSBox {
 
     addSubview(imageView)
 
-    imageView.fillColor = #colorLiteral(red: 0.847058823529, green: 0.847058823529, blue: 0.847058823529, alpha: 1)
+    // FIXME: Value of type 'NSImageView' has no member 'fillColor'
+//    imageView.fillColor = #colorLiteral(red: 0.847058823529, green: 0.847058823529, blue: 0.847058823529, alpha: 1)
```
jasonzurita: @dabbott Looks like there is some issue with generating `fillColor` here (run snapshot to see it), so I manually commented this line to allow the project to compile. 

Fixing this is outside of the scope of this PR.

dabbott: Good catch! Hmm... yeah, might have to get creative with NSImageView in that case. For another PR though.

examples/generated/test/LonaViewer/iOS/Generated.swift
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    var constraints: [Constraint] {
+        switch self {
+        case .localAsset:
+            return [
+                equal(\.topAnchor, \.safeAreaLayoutGuide.topAnchor),
+                equal(\.leftAnchor),
+            ]
+        case .pressableRootView:
```
dabbott: Can this case be combined with the previous?

jasonzurita: yeah, I will combine!

jasonzurita:
@dabbott 

1. Makes sense, I will move this to the `Lona/examples` directory.
2. The `Generated.swift` files are not auto-generated. I should have noted that. We definitely could make them auto-generated, but we would need to either:
i. Still add new Swift files in manually (existing generated Swift files will automatically update since they are already included in the Xcode project) or 
ii. Auto-add new Swift files to the Xcode `.xcodeproj` file, which would be interesting given the complexity of that file. One alternative to explore, would be to exclude the `xcodeproj` file from git and have the snapshot script also [build the Xcode project file using the Swift package manager](https://stackoverflow.com/questions/34116041/when-using-swift-package-manager-how-can-i-generate-an-xcode-project-file-for-de).

One additional thing to note about auto-generating the `Generated.swift` files, is that the generator would need to take into account the different auto layout constraints needed to properly test/use a given generated Swift file.

PR #179, opened 2018-08-30
Consolidate workspace browser and component editor

dabbott:
## What

Both the workspace browser and component editor now share a single window.

## Why

The UI is significantly less confusing.

## Testing Plan

cc: @outdooricon


ryngonzalez:
@dabbott Could you add screenshot for this change on here?

PR #183, opened 2018-09-04
Secondary button

outdooricon:
## What

I think it'd be good to get the secondaryButton example in now, before we start resolving. Just so we have a good baseline


outdooricon:
Great point about the order of precedence, i'll work with that in mind

PR #188, opened 2018-09-20
add instructions about npm install

mathieudutour:
## What

- [x] add instruction about running `npm install` with the correct node version

## Why

Otherwise the "export to sketch" does not work


studio/README.md
```diff
@@ -10,12 +10,15 @@ The easiest way to use Lona Studio is by downloading the prebuilt Mac App binary
...
 
-First, make sure you have [`bundler`](http://bundler.io/) and [Cocoapods](https://cocoapods.org/) installed. Then checkout the repo and run:
+First, make sure you have [`bundler`](http://bundler.io/), [Cocoapods](https://cocoapods.org/) and [nvm](https://github.com/creationix/nvm/#installation) installed. Then checkout the repo and run:
 
 ```
 cd studio
 bundle && bundle exec pod install
 carthage update --new-resolver --platform macOS
+cd LonaStudio/Scripting/Modules/component-to-sketch
+nvm use 8.4
```
ljharb: this can be reduced to just `nvm install 8.4`, and if you commit an `.nvmrc` file to the repo containing “8.4”, just `nvm install` will work.

dabbott:
I think I might put this in a subsection of `Building from Source` after the "If there are warnings (e.g. about project settings) you can ignore them." sentence. 

E.g.

```
### Sketch export (optional)

To enable Sketch export (beta), you'll also need to run:

cd...
nvm...
```

Since Sketch export is still pretty rough, and it's not strictly necessary for folks to run Lona Studio, I think it'd be better to call it out separately

dabbott:
Misc thoughts: eventually I want to bundle the Sketch stuff into a single .js file + native dependencies (if there are any) so that when people install the Lona Studio binary, it copies over to their Applications directory quickly. If the binary includes a node_modules directory, the copying takes forever.

mathieudutour:
Make sense. Just wanted to open this PR because I got a bit confused when the export didn't work

PR #196, opened 2018-09-21
Fix text style assignment in js logic

dabbott:
## What

Fix text style assignment in js logic

## Testing Plan

@mathieudutour

### DIFF
--- a/examples/generated/test/sketchJs/style/TextStyleConditional.js
+++ b/examples/generated/test/sketchJs/style/TextStyleConditional.js
@@ -16,7 +16,7 @@ export default class TextStyleConditional extends React.Component {
     }
     return (
       <View style={[ styles.view, {} ]}>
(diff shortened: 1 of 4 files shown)


examples/generated/test/sketchJs/style/TextStyleConditional.js
```diff
@@ -16,7 +16,7 @@ export default class TextStyleConditional extends React.Component {
     }
     return (
       <View style={[ styles.view, {} ]}>
-        <Text style={[ styles.text, { font: Text$textStyle } ]}>
+        <Text style={[ styles.text, { ...Text$textStyle } ]}>
```
mathieudutour: actually I'm wondering if `style={[ styles.text, Text$textStyle ]}` is enough. It should be, right?

dabbott: Yeah. Let's do that as an optimization though, cause I don't have time right now :P

PR #200, opened 2018-09-28
Allow JS properties to contain special chars, using bracket notation

outdooricon:
## What

I added the ability for JS to handle special chars in identifiers. I noticed this specifically in keys for colors, but really this could happen anywhere. We may still want to restrict the names of things from the LS perspective, but this will allow a little more flexibility for props.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -86,3 +86,4 @@ examples/generated/test/LonaViewer/LonaViewer.xcodeproj/project.xcworkspace/xcsh
 examples/LonaViewer/LonaViewer.xcodeproj/project.xcworkspace/contents.xcworkspacedata
 
 .vscode
+node_modules
--- a/compiler/core/src/javaScript/javaScriptRender.re
+++ b/compiler/core/src/javaScript/javaScriptRender.re
@@ -17,11 +17,17 @@ let renderBinaryOperator = x => {
   s(op);
 };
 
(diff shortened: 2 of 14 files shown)


.gitignore
```diff
@@ -86,3 +86,4 @@ examples/generated/test/LonaViewer/LonaViewer.xcodeproj/project.xcworkspace/xcsh
 examples/LonaViewer/LonaViewer.xcodeproj/project.xcworkspace/contents.xcworkspacedata
 
 .vscode
+node_modules
```
outdooricon: Something in vscode keeps creating this directory for me, with a debug log. So that's why i added it

outdooricon:
@mathieudutour brought up that swift offers something similar to JS's bracket notation that would fix the swift generated code in this example. I can back out the generated code changes in this PR if we're uncomfortable with having swift code like this until someone can change the swift compiler to do the same. Let me know what you think Devin

dabbott:
Since the Swift generation is normally working and usable, I don't think we should leave it in a broken state. 

For the Swift case I think we should strip out any `-` in identifiers, rather than doing the backtick notation (or whatever) since it's not idiomatic for this kind of thing.

dabbott:
So yeah, I guess just take out that color for now and the rest of the PR is fine

outdooricon:
Ok, test and generated code changes have been reverted

PR #201, opened 2018-09-28
Swift type generation

dabbott:
## What

Preliminary conversion from `types.json` to Swift types. No actual usage yet.

## Testing Plan

cc @outdooricon

### DIFF
--- a/compiler/core/src/main.re
+++ b/compiler/core/src/main.re
@@ -1,16 +1,16 @@
 open Node;
 
-[@bs.val] [@bs.module "fs-extra"] external ensureDirSync : string => unit = "";
+[@bs.val] [@bs.module "fs-extra"] external ensureDirSync: string => unit = "";
 
 [@bs.val] [@bs.module "fs-extra"]
-external copySync : (string, string) => unit = "";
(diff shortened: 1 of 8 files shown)


compiler/core/src/main.re
```diff
@@ -27,15 +27,20 @@ let options: LonaCompilerCore.Options.options = {
     | Some("airbnb") => Airbnb
     | _ => Standard
     },
-  filterComponents: getArgument("filterComponents")
+  filterComponents: getArgument("filterComponents"),
```
outdooricon: That's better :)

compiler/core/src/swift/swiftAst.re
```diff
@@ -127,15 +136,33 @@ and node =
...
       {
         .
         "name": string,
+        "isIndirect": bool,
         "inherits": list(typeAnnotation),
         "modifier": option(accessLevelModifier),
         "body": list(node),
       },
     )
+  | TypealiasDeclaration(
```
outdooricon: Should Alias be capitalized?

outdooricon: Ah got it

PR #203, opened 2018-10-01
Add Column Flex Direction as default

outdooricon:
## What

Web's flex direction default is `row`, but the expectation from lona is that `column` is the default. So, I'm adding column flexDirection alongside the flex declaration for react-dom.

## Why

Produce react-dom layouts consistent with LS.

## Major Changes

* Added flex direction default for react-dom
* Fixed the order that defaults are assigned to the ParameterMap, so that that explicit declarations are not overwritten by the defaults.

### DIFF
--- a/compiler/core/src/javaScript/javaScriptStyles.re
+++ b/compiler/core/src/javaScript/javaScriptStyles.re
@@ -66,6 +66,7 @@ let addDefaultStyles =
         empty
         |> add(ParameterKey.Display, LonaValue.string("flex"))
         |> add(ParameterKey.FlexDirection, LonaValue.string("column"))
+        |> add(ParameterKey.AlignItems, LonaValue.string("stretch"))
       )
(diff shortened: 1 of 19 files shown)


compiler/core/src/javaScript/javaScriptStyles.re
```diff
@@ -60,16 +60,16 @@ let addDefaultStyles =
...
     |> ParameterMap.filter((key, _) => Layer.parameterIsStyle(key));
   ParameterMap.assign(
-    styleParams,
     switch (framework) {
     | JavaScriptOptions.ReactDOM =>
-      ParameterMap.add(
-        ParameterKey.Display,
-        LonaValue.string("flex"),
-        ParameterMap.empty,
+      ParameterMap.(
```
dabbott: Nice use of inline module open

outdooricon:
Good idea, done!

PR #207, opened 2018-10-03
Shadows + Swift generation

dabbott:
## What

**Shadows**
- All views now support shadows
- Shadows can be used in logic
- Generated code is valid for iOS/macOS

**Compiler**
- All configs (colors, text styles, shadows) are passed around as a single variable `config`

screen shot 2018-10-02 at 7 13 35 pm

## Testing Plan

@outdooricon 
cc @ryngonzalez


dabbott:
Note that there are a couple quirks:
- In Lona Studio, only views with backgrounds will show shadows (this is due to how macOS/iOS renders things with shadow), which will be different than what is generated for web
- Similarly, in Lona Studio, a parent will clip its child's shadow
- In an existing workspace, you'll need to make a non-empty `shadows.json` before generating code. You can copy the one in this commit (`examples/test/shadows.json`)

outdooricon:
🕺

PR #209, opened 2018-10-05
Improve JS layout generation

dabbott:
## What

Components now expect their parent to use the platform default layout.

DOM:
```
display: flex
flex-direction: row
alignItems: flex-start
```

RN:
```
flex-direction: column
alignItems: stretch
```

This also fixes nested component layouts for React DOM.

## Why

Easier for consumers to use Lona components.

## Testing Plan

@outdooricon

### DIFF
--- a/examples/generated/test/react-dom/components/NestedButtons.js
+++ b/examples/generated/test/react-dom/components/NestedButtons.js
@@ -37,28 +37,20 @@ let styles = {
     display: "flex",
     flex: "1 1 0%",
     flexDirection: "column",
+    justifyContent: "flex-start",
     paddingTop: "24px",
     paddingRight: "24px",
     paddingBottom: "24px",
     paddingLeft: "24px"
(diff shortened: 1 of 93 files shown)


examples/generated/test/react-dom/components/NestedButtons.js
```diff
@@ -17,14 +17,14 @@ export default class NestedButtons extends React.Component {
...
       <ThemeProvider theme={theme}>
-        <div style={Object.assign(styles.view, {})}>
-          <Button style={Object.assign(styles.button, {})} label={"Button 1"} />
-          <div style={Object.assign(styles.view1, {})} />
-          <Button
-            style={Object.assign(styles.button2, {})}
-            label={"Button 2"}
-
-          />
+        <div style={Object.assign({}, styles.view, {})}>
```
outdooricon: 👏

PR #211, opened 2018-10-10
Support setting visibility in logic

dabbott:
## What

Generate JS and Swift based on visibility logic

## Testing Plan

cc @outdooricon

### DIFF
--- a/compiler/core/src/swift/swiftConstraint.re
+++ b/compiler/core/src/swift/swiftConstraint.re
@@ -251,12 +251,7 @@ let formatConstraintVariableName =
     ) => {
   open Constraint;
 
-  let verbose =
-    Sequence.occurrences(
-      a => Constraint.semanticEqual(a, const),
-      Constraint.allConstraints(combinations),
-    )
-    > 1;
+  let verbose = !Constraint.isAlwaysActivated(combinations, const);
 
   let formatAnchorVariableName = (layer: Types.layer, anchor) => {
     let anchorString = Constraint.anchorToString(anchor);
--- a/compiler/core/src/swift/swiftConstraint.re
+++ b/compiler/core/src/swift/swiftConstraint.re
@@ -244,15 +244,27 @@ let generateConstraintConstant =
 };
 
 let formatConstraintVariableName =
-    (rootLayer: Types.layer, constr: Constraint.t, verbose: bool) => {
+    (
+      combinations: list(Constraint.visibilityCombination),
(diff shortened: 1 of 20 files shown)


compiler/core/src/swift/swiftConstraint.re
```diff
@@ -380,22 +434,175 @@ let setUpFunction =
...
-              ...constraints |> List.map(assignConstraintIdentifier),
+              ...allConstraints |> List.map(assignConstraintIdentifier),
             ] :
             [],
         ],
       ),
   });
+};
+
+/* private func conditionalConstraints() -> [NSLayoutConstraint] {
```
outdooricon: Did you still want this?

compiler/core/src/utils/constraint.re
```diff
@@ -379,4 +430,89 @@ let getConstraints =
...
+
+         {
+           rootLayer,
+           visibleLayers,
+           constraints: getConstraints(getComponent, rootLayer),
+           variableName: "",
+         };
+       });
+
+  /* Js.log3(
```
outdooricon: same here

examples/generated/test/react-dom/style/VisibilityTest.js
```diff
@@ -0,0 +1,69 @@
...
+    let theme = {
+      "container": { "normal": {} },
+      "inner": { "normal": {} },
+      "title": { "normal": {} },
+      "view": { "normal": {} }
+    }
+    return (
+      <ThemeProvider theme={theme}>
+        <div style={Object.assign({}, styles.container, {})}>
+          {false && <div style={Object.assign({}, styles.inner, {})} />}
```
outdooricon: Is this just the resulting of assigning visibility straight as a lone prop instead of assigning through conditional?

compiler/core/src/swift/swiftConstraint.re
dabbott: I think it's kind of convenient but maybe I'll clean it up a little and add some notes

compiler/core/src/utils/constraint.re
dabbott: I'll remove this one

examples/generated/test/react-dom/style/VisibilityTest.js
dabbott: yeah. I didn't optimize this case at all since really you should just delete it from the Lona component file... but it'll at least look correct if you don't

PR #212, opened 2018-10-11
Add shadows for reactDom

outdooricon:
## What

I added shadows for react dom. It doesn't generate any shadows in the shadows file for the other JS frameworks right now. I believe that shadowsTest.components will continue to be unusable for react-native and sketch. 

I also added watchers for the different compiler snapshot targets, using nodemon. It gives almost instant feedback as you modify the compiler, so I've found it really helpful and thought it should be shared.

### DIFF
--- a/examples/generated/test/react-dom/shadows.js
+++ b/examples/generated/test/react-dom/shadows.js
@@ -1,0 +1,7 @@
+import colors from "./colors"
+
+export default {
+  elevation1: { filter: "drop-shadow(0px 1px 2px rgba(0,0,0,0.5));" },
+  elevation2: { filter: "drop-shadow(0px 2px 4px rgba(0,0,0,0.5));" },
+  elevation3: { filter: "drop-shadow(0px 3px 9px " + colors.grey900 + ");" }
+};
--- a/examples/generated/test/swift/Shadows.swift
+++ b/examples/generated/test/swift/Shadows.swift
@@ -3,5 +3,5 @@ import UIKit
 public enum Shadows {
   public static let elevation1 = Shadow(color: #colorLiteral(red: 0, green: 0, blue: 0, alpha: 0.5), offset: CGSize(width: 0, height: 1), blur: 2)
   public static let elevation2 = Shadow(color: #colorLiteral(red: 0, green: 0, blue: 0, alpha: 0.5), offset: CGSize(width: 0, height: 2), blur: 4)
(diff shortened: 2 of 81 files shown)


examples/generated/test/swift/Shadows.swift
```diff
@@ -3,5 +3,5 @@ import UIKit
 public enum Shadows {
   public static let elevation1 = Shadow(color: #colorLiteral(red: 0, green: 0, blue: 0, alpha: 0.5), offset: CGSize(width: 0, height: 1), blur: 2)
   public static let elevation2 = Shadow(color: #colorLiteral(red: 0, green: 0, blue: 0, alpha: 0.5), offset: CGSize(width: 0, height: 2), blur: 4)
-  public static let elevation3 = Shadow(color: #colorLiteral(red: 0, green: 0, blue: 0, alpha: 0.5), offset: CGSize(width: 0, height: 3), blur: 9)
+  public static let elevation3 = Shadow(color: Colors.grey900, offset: CGSize(width: 0, height: 3), blur: 9)
```
dabbott: Good thinking! I totally forgot about using system colors here

examples/generated/test/react-dom/shadows.js
```diff
@@ -0,0 +1,7 @@
+import colors from "./colors"
+
+export default {
+  elevation1: { filter: "drop-shadow(0px 1px 2px rgba(0,0,0,0.5));" },
+  elevation2: { filter: "drop-shadow(0px 2px 4px rgba(0,0,0,0.5));" },
+  elevation3: { filter: "drop-shadow(0px 3px 9px " + colors.grey900 + ");" }
```
dabbott: This object format seems like the best approach, since at some point we could add a flag to use box-shadow or drop-shadow, and we wouldn't have to change anything elsewhere.

The other option would be to use just the string:

`elevation1: "drop-shadow(0px 1px 2px rgba(0,0,0,0.5));"`

And then in the JSX, `filter: shadows.elevation1`. That might sometimes result in one less object spread, but other than that I don't see any advantage.

Are there any other tradeoffs? How did you choose this format?

outdooricon: I went this direction just in case we had to augment the css with more than just filter to reproduce the right shadow. It felt the most flexible, but I'm open to changing it if you'd like

dabbott: I think this is good. Let's go with this, and we can always revisit if a new use case or something comes up.

PR #213, opened 2018-10-12
Optional component parameters

dabbott:
## What

Optional component parameters now generate correct code for JS and Swift. When used in logic, they can be unwrapped with the `If let` statement.

Note: I'm not sure if optional custom types work -- they probably don't, but we can add that separately if needed. Default values for parameters don't do anything yet.

## Testing Plan

@outdooricon

### DIFF
--- a/examples/generated/test/react-native/logic/Optionals.js
+++ b/examples/generated/test/react-native/logic/Optionals.js
@@ -9,8 +9,11 @@ export default class Optionals extends React.Component {
   render() {
 
     let Label$text
+    let StringParam$text
     let View$backgroundColor
+    let unwrapped
     Label$text = ""
+    StringParam$text = "No string param"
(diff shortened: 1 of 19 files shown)


examples/generated/test/react-native/logic/Optionals.js
```diff
@@ -0,0 +1,70 @@
...
+
+    if (this.props.boolParam == true) {
+      Label$text = "boolParam is true"
+      View$backgroundColor = colors.green200
+    }
+    if (this.props.boolParam == false) {
+      Label$text = "boolParam is false"
+      View$backgroundColor = colors.red200
+    }
+    if (this.props.boolParam == null) {
```
dabbott: Using `==` instead of `===` here makes it so that `undefined` will work too, for when a component is consumed outside of Lona and _no props_ is passed.

PR #214, opened 2018-10-12
Regenerate the Lona workspace used by LS

dabbott:
## What

Regenerate all swift files used by LS after lots of compiler updates

## Testing Plan


dabbott:
We shouldn't be linted generated files... but will ignore for now

PR #232, opened 2018-10-23
Backing element customization

dabbott:
## What

Support changing the underlying React element generated for any layer

## Testing Plan

cc @outdooricon


outdooricon:
I see a bright future with `metadata`

PR #237, opened 2018-10-25
Vector layers within the same file share an underlying component

dabbott:
## What

Fixes an issue where a component using the same svg twice generates invalid code.

## Testing Plan

@outdooricon


outdooricon:
👍 perfect

PR #241, opened 2018-10-31
Turn numberOflines into maxheight for react-dom

outdooricon:
## What

Since react-dom doesn't know what to do with numberOfLines, for now we're going to use max-height to cut off the remaining text. Maxheight is calculated off of the line-height of the text style.

### DIFF
--- a/compiler/core/src/javaScript/javaScriptStyles.re
+++ b/compiler/core/src/javaScript/javaScriptStyles.re
@@ -376,7 +376,7 @@ let handleNumberOfLines =
     ParameterMap.find_opt(ParameterKey.NumberOfLines, parameters),
   ) {
   | (JavaScriptOptions.ReactDOM, Some(lineCount)) =>
-    open Monads;
+    open Monad;
     let lineHeight =
       switch (
         ParameterMap.find_opt(ParameterKey.TextStyle, parameters)
--- a/compiler/core/src/javaScript/javaScriptStyles.re
+++ b/compiler/core/src/javaScript/javaScriptStyles.re
@@ -369,6 +369,44 @@ let getStylePropertyWithUnits =
   | (_, _) => getStyleProperty(framework, key, colors, value)
   };
 
+let handleNumberOfLines =
+    (framework: JavaScriptOptions.framework, config: Config.t, parameters) =>
+  switch (
+    framework,
+    ParameterMap.find_opt(ParameterKey.NumberOfLines, parameters),
+  ) {
+  | (JavaScriptOptions.ReactDOM, Some(lineCount)) =>
+    open Monads;
+    let lineHeight =
+      switch (
+        ParameterMap.find_opt(ParameterKey.TextStyle, parameters)
+        >>= (
+          (item: Types.lonaValue) =>
+            TextStyle.find(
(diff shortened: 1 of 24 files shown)


compiler/core/src/javaScript/javaScriptStyles.re
```diff
@@ -369,6 +369,44 @@ let getStylePropertyWithUnits =
...
+      ) {
+      | Some(lineHeight) => lineHeight
+      | None => 0.0
+      };
+
+    parameters
+    |> ParameterMap.remove(ParameterKey.NumberOfLines)
+    |> ParameterMap.add(
+         ParameterKey.Display,
+         LonaValue.string("inline-block"),
```
dabbott: Should text elements be `block` for now? Thinking about it some more, we don't really support any inline elements yet. So `inline-block` is essentially just `block` but quirkier here. Same goes for other text, even when max-lines case isn't involved. I.e. we could just add `display: block` as a default on Text.


In the future I've been considering having two separate kinds of `Text` layer. One for "block" text that can contain multiple inline elements (e.g. text with different styles, images) and one for "inline" text, that goes within a "block" text (or is auto-wrapped at compile-time).

compiler/core/src/utils/monads.re
```diff
@@ -0,0 +1,7 @@
+let bindOption = (f: 'a => option('b), value: option('a)): option('b) =>
```
dabbott: Let's just call this file `monad.re` to be slightly more consistent with haskell's naming (theirs is `import Control.Monad`)

compiler/core/src/javaScript/javaScriptStyles.re
```diff
@@ -369,6 +369,44 @@ let getStylePropertyWithUnits =
...
+  switch (
+    framework,
+    ParameterMap.find_opt(ParameterKey.NumberOfLines, parameters),
+  ) {
+  | (JavaScriptOptions.ReactDOM, Some(lineCount)) =>
+    open Monads;
+    let lineHeight =
+      switch (
+        ParameterMap.find_opt(ParameterKey.TextStyle, parameters)
+        >>= (
```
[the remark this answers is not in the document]
outdooricon: i brought it back after i got it to pipe... whereas before, it was needing to be nested

compiler/core/src/utils/monads.re
```diff
@@ -0,0 +1,7 @@
...
+  switch (value) {
+  | None => None
+  | Some(a) => f(a)
+  };
+
+let (>>=) = (x, y) => x |> (y |> bindOption);
```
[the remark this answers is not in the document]
outdooricon: 😭

compiler/core/src/javaScript/javaScriptStyles.re
outdooricon: i like the default approach 👍

outdooricon: wow, I need my coffee... sigh... I'll add soon :)

outdooricon: ok, resolved. I also brought in the default text-align we talked about

dabbott:
LGTM!

PR #248, opened 2018-11-08
Add Unit-ed values to JSX Styling

outdooricon:
## What

When styles are generated, they look for framework specific units. However, there was nothing in the JSX styling that did the same. This adds that.

### DIFF
--- a/compiler/core/src/javaScript/utils/reactTranslators.re
+++ b/compiler/core/src/javaScript/utils/reactTranslators.re
@@ -23,7 +23,7 @@ let convertUnitlessAstNode =
     switch (value) {
     | Ast.Identifier(path) =>
       Ast.BinaryExpression({
-        left: Ast.Identifier(path),
+        left: value,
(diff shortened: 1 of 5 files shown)


compiler/core/src/javaScript/utils/reactTranslators.re
```diff
@@ -0,0 +1,39 @@
...
+  };
+
+let convertUnitlessAstNode =
+    (framework: JavaScriptOptions.framework, value: Ast.node) =>
+  switch (framework) {
+  | JavaScriptOptions.ReactDOM =>
+    switch (value) {
+    | Ast.Identifier(path) =>
+      Ast.BinaryExpression({
+        left: Ast.Identifier(path),
```
dabbott:
```suggestion
        left: value,
```

outdooricon:
Ok, I'm not entirely clear on how to do that but we can just chat more specifically about it when I encounter it again

PR #250, opened 2018-11-13
Styled components

dabbott:
## What

Support generating React DOM code using styled-components for styling. In theory RN and Sketch should work too, though I haven't tested.

This can be configured with the `--styleFramework=styledcomponents` CLI flag.

A couple things to note:
- Dynamic styles (set via logic) are still applied using the `style` jsx attribute of the element. I'm not sure if styled-components prefixes these -- but all of the styles we support setting dynamically should work on every browser without vendor prefixing, except maybe shadow.
- I temporarily removed an svg optimization where I reused the same svg component if the svg is used twice within the same file. I'm not sure what's the best way to support this using styled-components.
- A left a couple theme-related things commented since I don't want to forget what we had planned.
- For RN/Sketch there will be naming collisions if you leave your layers as default names like "View", but we can deal with that if we ever end up using styled-components with those.

## Testing Plan

@outdooricon

### DIFF
--- a/compiler/core/package.json
+++ b/compiler/core/package.json
@@ -4,11 +4,7 @@
   "description": "Lona cross-platform code compiler",
   "main": "build/index.js",
   "bin": "./build/index.js",
-  "files": [
-    "build/index.js",
-    "static",
-    "README.md"
-  ],
+  "files": ["build/index.js", "static", "README.md"],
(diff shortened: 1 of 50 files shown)


ryngonzalez:
WOOOOOO

PR #253, opened 2018-11-14
Used backingElementClass with styled-components

dabbott:
## What

Use the backingElementClass when creating styled-components

## Testing Plan

cc @outdooricon


outdooricon:
💥 thank you!

PR #268, opened 2018-12-15
Fix memory leaks, remove legacy type and some yoga layouts

dabbott:
## What

This fixes the main memory leak caused by showing or modifying the logic tab. As part of this process, I also removed the legacy Enumeration type and some usage of yoga layouts in favor of NSStackView.

## Testing Plan

I tested the main parameters/logic flows, but there may still be crashes introduced by this. Let me know if you encounter any logic-related crashes.

@outdooricon @ryngonzalez


outdooricon:
Yes!

PR #272, opened 2019-01-03
Fix issue where utility tabs don't show/hide correctly

dabbott:
## What

I'm not sure why yet but `view.superview` isn't `self`. In any case, the check is extraneous, so I'm removing it.

## Testing Plan

cc @outdooricon


outdooricon:
so much ❤️

PR #275, opened 2019-01-05
image viewer and a few other fixes

mathieudutour:
## What

I've been playing around the codebase, trying to get familiar with it and did a few things:
- clicking on an image will show it: 
screenshot 2019-01-05 at 16 05 25

- fix a few typos
- make sure that when clicking on a folder, it doesn't print a bunch of errors
- when saving the text styles, trigger the right event
- add a specific icon for lona.json

## Feedback

The generated `ImageViewer.swift` contains 3 lines to position the image view by setting properties that don't exist on NSImageView. I'm not sure what's the way to work around that

cc: @dabbott


dabbott:
Wow, this is awesome! I'd been thinking an image viewer would be convenient, and now we have one! I'm gonna try to look now, but if I run out of time, then later today.

Btw, the tests spontaneously broke about 2 days ago, I think due to a ruby version issue, and I was hoping they'd spontaneously fix themselves. It looks like that's not gonna happen though, so I'll fix em soon. Anyway, when this is ready I'll just force merge it -- the tests only check Swiftlint for the most part and aren't essential.

dabbott:
Ok! I pulled this down and tried it out. Overall it looks really good. We'll probably just want to change the way the image in the image viewer works a bit.

It's interesting to see how you ended up getting the image to work, and actually I think this is pretty close to what we want here. Using `Backing Element` and `types.json` are both escape hatches for when there isn't a built-in type/component for what you need. I think we do want a `Backing Element`, but for a custom image class (I'll explain below) and then we can use a built-in type instead of adding to `types.json`. There are 3 main changes to make:

1. For images, we should start with the built-in `Image` component, rather than using a `View` and setting the `Backing Element` to `NSImageView`. You'll have to delete and re-create this `View` using an `Image`:

    screen shot 2019-01-05 at 5 08 01 pm

    I'm kind of surprised Lona Studio let you assign to the `.image` parameter of a `View`, so I'm wondering if there was a bug and you _did_ make an `Image` but Lona forgot somehow and turned it into a `View`.

2. Anyway, after that, we'll want a parameter thats a `URL` that is `optional`.

    screen shot 2019-01-05 at 5 09 33 pm

    Type doesn't _have_ to be optional, since it's fine to construct an empty `NSImage()` as the default value, but that comes down to preference. We won't need to use `types.json` here.

    > In Lona, the `URL` type is used for images... and should potentially be renamed to `Image`.

3. From there, we need to figure out how the image should be displayed. Lona offers 3 built-in resizing modes for displaying an image:

    screen shot 2019-01-05 at 5 35 17 pm

    Probably aspect-preserving fit is the best choice, but that's not ideal. If the image is really small, it'll be needlessly scaled up and look bad. 

    The way it is now uses the image's intrinsic dimensions, which has sort of the reverse problem: a large image will expand the window to infinity:

    screen shot 2019-01-05 at 4 57 14 pm

    That's actually one of the reasons why I don't allow using the image's intrinsic dimensions in Lona at the moment (the other relates to supporting this on web). Using an image's intrinsic dimensions directly is rarely the right algorithm for arbitrary images.

    In this case we'd probably want a combination of the two, like... if the image size is less than the viewer size, use the intrinsic dimensions, otherwise use aspect-preserving fit. But since there isn't a built-in way to do that, I'd say the easiest way is to use a `Backing Element` after all.

    I copied & pasted the `LNAImageView` (used for images when generating macOS code, to support content modes like iOS), tweaked the `viewWillDraw` to be the behavior we want, and renamed it `ImagePreview`:

    ```swift
    public class ImagePreview: NSImageView {
        // ... the rest of the contents of LNAImageView

        override public func viewWillDraw() {
            if let image = image, let originalImage = originalImage {
                if originalImage.size.width <= frame.width && originalImage.size.height <= frame.height {

                } else if image.size != originalImage.size.resized(within: bounds.size, usingResizingMode: resizingMode).size {
                    super.image = originalImage.resized(within: bounds.size, usingCroppingMode: resizingMode)
                }
            }

            super.viewWillDraw()
        }
    }
    ```

    This will be useful in other places in Lona Studio, so I think it's a reasonable addition. It could be a subclass of `LNAImageView`, but I don't want to have to change the private field `originalImage` to internal at the moment.

    Lastly, set the backing element of the `Image` element to the new `ImagePreview` class.

I wasn't 100% confident about the custom image resizing algo, so I ended up making these changes... and it seems to work as expected. So you can either make them yourself for practice, or merge my branch `https://github.com/airbnb/Lona/tree/image-previewer` into yours. 

Misc nits (which I didn't do):
- remove the `ImageDataOrNil` from `types.json`
- probably rename `imageData` to `image`, now that we're using the built-in type
- the .lona file icon is a pixel too far to the right I think, compared with the others (I should commit the Sketch file for these). 

Other than that, the rest LGTM. The list index bugfix is great, I think I just introduced this, but didn't have time to look at it yet.

mathieudutour:
Thanks for the in-depth review!

> I'm kind of surprised Lona Studio let you assign to the `.image` parameter of a `View`

It did not 😆. The bottom panel doesn't show as soon as the component is a bit complicated. Only the "tab bar" shows up, regardless of if I click on the icon which I assume is to toggle it on the top right. So I modified the `.component` manually.

dabbott:
>  The bottom panel doesn't show as soon as the component is a bit complicated. Only the "tab bar" shows up, regardless of if I click on the icon which I assume is to toggle it on the top right. 

Oooh. I never implemented the toggle icon 😂, but you can click and drag the bottom bar up. Something to fix...

dabbott:
I'm gonna try to fix the build now

PR #280, opened 2019-01-07
Allow workspace name customization

dabbott:
## What

The workspace name is customizable via preferences. Requires hit-enter-to-confirm.

This should fix the bugs around preferences saving to the wrong workspace, and the preferences UI loading twice (though not 100% sure).

cc @outdooricon


outdooricon:
Thank you Devin, this is perfect!

PR #282, opened 2019-01-09
polish folder view

mathieudutour:
## What

- [x] reopen the workspace when quitting it from a folder view and reopening Lona (previously it wouldn't cause Lona don't know how to open a folder)
- [x] do not show the welcome window when a workspace is being reopened
- [x] show the folder view for the root when opening a workspace

cc: @dabbott


mathieudutour:
Sounds good 👌

PR #284, opened 2019-01-10
update danger to remove deprecated danger-swiftlint

mathieudutour:
## What

- [x] update danger to remove deprecated danger-swiftlint

cc: @dabbott


dabbott:
@mathieudutour should I be looking at this? It has [WIP] so I ignored it

mathieudutour:
Not yet, for some reason it’s timing out on travis but works fine locally. Haven’t have time to look more into it

dabbott:
Ah ok. I restarted the travis build. The build times out like 10% of the time, I've never looked at why :/

PR #288, opened 2019-01-20
Lona web accessibility support

dabbott:
## What

This adds initial accessibility support to JS DOM components. The doc `studio/docs/accessibility.md` explains how everything works.

There's still more to do, particularly around mapping roles over correctly, and adding more roles for web.

cc @outdooricon


outdooricon:
👀

examples/generated/test/react-dom/components/ComponentParameterInstance.js
```diff
@@ -10,6 +10,7 @@ export default class ComponentParameterInstance extends React.Component {
   render() {
 
 
+
```
outdooricon: this is odd... i wonder why we're getting another newline now in so many places... is there an optional that defaults to newline for `None`, that's getting hit a lot more now?

examples/generated/test/react-dom/interactivity/AccessibilityTest.js
```diff
@@ -4,10 +4,72 @@ import styled from "styled-components"
...
+
+    if (nextIndex >= focusElements.length) {
+      this.props.onFocusNext && this.props.onFocusNext()
+      return ;
+    }
+
+    focusElements[nextIndex].focus && focusElements[nextIndex].focus()
+  }
+
+  focusPrevious = ({ focusRing = true } = { focusRing: true }) => {
```
outdooricon: I think we should probably declare now that people will need to have a way to transform class properties in their consuming app.

examples/generated/test/react-dom/interactivity/AccessibilityTest.js
```diff
@@ -71,8 +162,9 @@ let CheckboxRow = styled.div({
   paddingTop: "10px",
   paddingRight: "10px",
   paddingBottom: "10px",
-  paddingLeft: "10px"
-})
+  paddingLeft: "10px",
+  ...!props.focusRing && { ":focus": { outline: 0 } }
```
outdooricon: this looks like the beginning of the future!

compiler/core/package.json
```diff
@@ -4,36 +4,37 @@
   "description": "Lona cross-platform code compiler",
   "main": "build/index.js",
   "bin": "./build/index.js",
-  "files": [
-    "build/index.js",
-    "static",
-    "README.md"
-  ],
+  "files": ["build/index.js", "static", "README.md"],
```
outdooricon: It looks like prettier has been in here ㊙️

compiler/core/src/javaScript/javaScriptRender.re
```diff
@@ -127,26 +136,33 @@ let rec render = ast: Prettier.Doc.t('a) =>
...
   | ArrowFunctionExpression(o) =>
-    let parameterList = o.params |> List.map(s) |> join(line);
+    let parameterList = o.params |> List.map(render) |> join(line);
 
     switch (o.body) {
     | [Return(ObjectLiteral(_) as literal)] =>
       group(s("(") <+> parameterList <+> s(") => ("))
       <+> render(literal)
       <+> s(")")
+    | [Return(inner)] =>
```
outdooricon: loving this update!

compiler/core/src/javaScript/javaScriptAst.re
```diff
@@ -57,13 +58,19 @@ and binaryExpression = {
...
 }
+and unaryExpression = {
+  prefix: bool,
+  operator: string,
+  argument: node,
+}
 and ifStatement = {
   test: node,
   consequent: list(node),
+  alternate: list(node),
```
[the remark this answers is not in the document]
dabbott: Yep. I'm using the babylon names just to have something to follow, and they're a bit weird.

compiler/core/src/javaScript/javaScriptAst.re
```diff
@@ -57,13 +58,19 @@ and binaryExpression = {
   operator: binaryOperator,
   right: node,
 }
+and unaryExpression = {
```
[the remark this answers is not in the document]
dabbott: Yes lol, though I think when we turn floats to strings they have the minus sign built in

examples/generated/test/react-dom/components/ComponentParameterInstance.js
dabbott: I previously had a special case in one of the list output functions in `javaScriptRender` that wasn't quite correct, but I left it for a while to reduce the extra newlines. I'll take a look at these in another PR

examples/generated/test/react-dom/interactivity/AccessibilityTest.js
dabbott: Good thinking. I think it's OK since anybody using JSX likely should be set up for this, but we can also go with a constructor at some point if needed

PR #289, opened 2019-01-21
Add borderStyle support for JS

outdooricon:
## What

This adds BorderStyle support for the JS compilers. I'm sure the output for swift is currently incorrect, and I'm not entirely sure what to do with it. @dabbott do you want me to try to solve for it? Or do you want to take it over when you add the LS side?

### DIFF
--- a/compiler/core/src/core/layer.re
+++ b/compiler/core/src/core/layer.re
@@ -60,6 +60,7 @@ let getParameterCategory = (x: ParameterKey.t) =>
   | Opacity => Style
   | BorderColor => Style
   | BorderRadius => Style
+  | BorderStyle => Style
   | BorderWidth => Style
   | TextAlign => Style
   | Shadow => Style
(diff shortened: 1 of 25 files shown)


dabbott:
Is this how it works: if there's no `borderStyle` in component's params, none will be printed, unless necessary for logic?

I can handle the Swift part. It doesn't translate directly. For this PR, let's just turn anything with borderStyle to an empty node. Basically how I set the AppKit version of `accessibilityType` to empty here: https://github.com/airbnb/Lona/blob/master/compiler/core/src/swift/swiftLogic.re#L209-L232

outdooricon:
That's right, if there's no borderStyle it doesn't add one. This is for the use case example of applying a border radius without needing a border. I had imagined the possibility that LS wouldn't even allow you to add a border width unless you had first selected a border style other than none

PR #300, opened 2019-01-29
update to react-sketchapp@3

mathieudutour:
## What

- [x] remove dependency on react-sketchapp fork
- [x] use `sketch-file` to generate the sketch file

cc: @dabbott


mathieudutour:
> Also, you probably already know this, but LS doesn't output the flexToSketchJSON stuff anymore, so this code would need to be changed a bit to go through the compiler before it's usable again.

yeah we probably need to use the compiler directly

mathieudutour:
ah yeah it's pretty old I think

PR #302, opened 2019-01-30
webpack loader to compile .component directly

mathieudutour:
## What

- [x] load `.component` files directly from webpack
- [x] handle json files
- [x] write doc


compiler/webpack-loader/lib/loader.js
```diff
@@ -0,0 +1,106 @@
+const path = require('path')
+const fs = require('fs')
+const qs = require('querystring')
+const { getOptions } = require('loader-utils')
+const { exec } = require('child_process')
+
+const IMPORT_REGEX = /import [a-zA-Z]+ from "([a-zA-Z./]+)"/g
```
dabbott: Depending on the length of the imports, there might be newlines between the different parts of this, so we might want to match arbitrary whitespace.

In some rare cases I think we might actually import things other than the default component, e.g. `import Button, { BUTTON_TYPES } from...`. But we can handle that later after we graduate from regex.

mathieudutour: Yeah this is just a hack until we get dependencies info from the lona compiler directly

mathieudutour:
btw, the fact that `colors.json` (and other similar files) is a json file means that the loader can't know that it should be compiled by lona.

I work around this by forcing the compilation when importing a component when it means that if you import the colors file directly (maybe with a custom component), you will just get the json, not the compiled js file.

Maybe that's what we want (if you import a json file then you get a json file) but if we'd like to get the compiled file, then I see 2 solutions:
- custom extension like for the components
- adding `"lona": true," at the start of the JSON so that when the loader gets a json file, it can look if there is this field and compile it.

Both aren't ideal tho :/

dabbott:
Another sort of in-between option would be to go with `core.colors.json`, `core.textStyles.json`. We'd be able to identify the files as Lona JSON, and the files would still open in text editors as JSON. One advantage of this or a custom extension is we open up the path for having multiple e.g. sets of color files in the future, `dark-theme.colors.json`, `brand.colors.json`. Are there downsides to this option?

Possibly we could also check for a `lona: true` if we support reading styled-dictionary files at some point and want easier interop.

mathieudutour:
yeah `.colors.json` sounds good

PR #303, opened 2019-02-04
allow having common plugins stored in the app support folder

mathieudutour:
## What

The plugins are now sourced from both
- the `/plugins` folder of the current workspace
- the `~/Library/Application Support/${BUNDLE_IDENFIFIER}/plugins` folder

cc: @dabbott

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ env:
 before_install:
   - brew update
   - brew outdated carthage || brew upgrade carthage
+  - brew tap danger/tap
   - brew install danger/tap/danger-swift
   - BUNDLE_GEMFILE=studio/Gemfile bundle install
   - BUNDLE_GEMFILE=studio/Gemfile bundle exec pod repo update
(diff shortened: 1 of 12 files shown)


dabbott:
Tests fail with:

```
Error: No available formula with the name "danger/tap/danger-swift" 
Please tap it and then try again: brew tap danger/tap
==> Searching for similarly named formulae...
Error: No similarly named formulae found.
The command "brew install danger/tap/danger-swift" failed and exited with 1 during .
```

I forget, did we change this recently?

dabbott:
Alright, let's add it and see if we use it.

Any idea on the tests? I sorta remember changing the danger stuff recently

mathieudutour:
I'm not sure, there is def `danger/tap/danger-swift` available for me but homebrew was updated to 2.0 yesterday so it might be something related

dabbott:
Oh interesting, that must be it then. I guess with 2.0 you have to run `tap` before `install`. I'm do that now

dabbott:
CI is fixed, can you rebase this on master?

PR #306, opened 2019-02-06
add POC of rpc for plugins

mathieudutour:
## What

- [x] add RPC for plugins
- [x] create companion library: https://github.com/mathieudutour/stdio-jsonrpc

That's what a plugin looks like:
```js
const { setup, sendRequest, sendNotification, RPCError } = require('stdio-jsonrpc')

setup({
  onRequest(method, params, callback) {
    console.log(method, params)
    callback(new RPCError.MethodNotFound())
  },
  onNotification(method, params) {
    console.log(method, params)
  }
})

sendRequest('workspace_path')
  .then(workspace => sendNotification('alert', {msg: workspace}))
  .catch(x => console.error(x))
  .then(() => process.exit(0))

```

cc: @dabbott

### DIFF
--- a/studio/LonaStudio/Plugins/LonaPlugins.swift
+++ b/studio/LonaStudio/Plugins/LonaPlugins.swift
@@ -48,7 +48,7 @@ class LonaPlugins {
 
             let rpcService = RPCService()
 
-            let sendData = LonaNode.run(
+            let sendData = LonaNode.launch(
                 arguments: [url.appendingPathComponent(config.main).path],
                 currentDirectoryPath: url.path,
                 onData: rpcService.handleData,
--- a/studio/LonaStudio/Plugins/LonaPlugins.swift
+++ b/studio/LonaStudio/Plugins/LonaPlugins.swift
@@ -48,10 +48,10 @@ class LonaPlugins {
 
             let rpcService = RPCService()
 
-            let stdinFileHandle = LonaNode.run(
+            let sendData = LonaNode.run(
                 arguments: [url.appendingPathComponent(config.main).path],
                 currentDirectoryPath: url.path,
-                onData: rpcService.handleRaw,
+                onData: rpcService.handleData,
                 onSuccess: { output in
                     Swift.print("Output", String(data: output, encoding: String.Encoding.utf8) ?? "")
 
@@ -63,7 +63,7 @@ class LonaPlugins {
 //                    }
             })
 
-            rpcService.sendData = stdinFileHandle?.write
+            rpcService.sendData = sendData
         }
 
         // MARK: Private
--- a/studio/LonaStudio/Plugins/LonaPlugins.swift
+++ b/studio/LonaStudio/Plugins/LonaPlugins.swift
@@ -46,11 +46,14 @@ class LonaPlugins {
         func run(onSuccess: (String) -> Void) {
             guard let config = config else { return }
 
-            LonaNode.run(
+            let rpcService = RPCService()
+
+            let stdinFileHandle = LonaNode.run(
                 arguments: [url.appendingPathComponent(config.main).path],
                 currentDirectoryPath: url.path,
+                onData: rpcService.handleRaw,
                 onSuccess: { output in
-                    Swift.print("Output", output.utf8String() ?? "")
+                    Swift.print("Output", String(data: output, encoding: String.Encoding.utf8) ?? "")
 
 //                    DispatchQueue.main.async {
 //                        let alert = NSAlert()
@@ -59,6 +62,8 @@ class LonaPlugins {
 //                        alert.runModal()
 //                    }
             })
+
+            rpcService.sendData = stdinFileHandle?.write
         }
 
         // MARK: Private
--- a/studio/LonaStudio/Plugins/RPCService.swift
+++ b/studio/LonaStudio/Plugins/RPCService.swift
@@ -1,0 +1,184 @@
+//
+//  RPCService.swift
+//  LonaStudio
+//
+//  Created by Mathieu Dutour on 06/02/2019.
+//  Copyright © 2019 Devin Abbott. All rights reserved.
+//
+
+// most of it is taken from https://github.com/xi-editor/xi-mac/blob/master/Sources/XiEditor/RPCSending.swift
+
+import Foundation
+import AppKit
+
+// An error returned from core
+struct RemoteError {
+    let code: Int
+    let message: String
+    let data: AnyObject?
+
+    init?(json: [String: AnyObject]) {
+        guard let code = json["code"] as? Int,
+            let message = json["message"] as? String else { return nil }
+        self.code = code
+        self.message = message
+        self.data = json["data"]
+    }
+}
+
+// The return value of a synchronous RPC
+enum RpcResult {
+    case error(RemoteError)
+    case ok(AnyObject)
+}
+
+enum RPCRequestMethod: String {
+    case workspacePath = "workspace_path"
+}
+
+enum RPCNotificationMethod: String {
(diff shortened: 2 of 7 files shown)


studio/LonaStudio/Plugins/RPCService.swift
```diff
@@ -0,0 +1,184 @@
...
+}
+
+// The return value of a synchronous RPC
+enum RpcResult {
+    case error(RemoteError)
+    case ok(AnyObject)
+}
+
+enum RPCRequestMethod: String {
+    case workspacePath = "workspace_path"
```
dabbott: Lona uses camelCase everywhere I think, so let's go with `"workspacePath"`.

studio/LonaStudio/Plugins/LonaPlugins.swift
```diff
@@ -59,6 +62,8 @@ class LonaPlugins {
 //                        alert.runModal()
 //                    }
             })
+
+            rpcService.sendData = stdinFileHandle?.write
```
dabbott: Hmm... does this prevent `rpcService` from getting garbage-collected?

studio/LonaStudio/Plugins/RPCService.swift
```diff
@@ -0,0 +1,184 @@
...
+              let method = RPCRequestMethod(rawValue: jsonMethod) else {
+                sendError(id: id, error: [
+                    "code": -32601,
+                    "message": "Method not found"
+                ])
+                return
+        }
+
+        switch method {
+        case .workspacePath:
```
dabbott: Let's move our own code out of the Xi stuff so it's easier to update their code if needed

studio/LonaStudio/Plugins/LonaPlugins.swift
```diff
@@ -46,11 +46,14 @@ class LonaPlugins {
         func run(onSuccess: (String) -> Void) {
             guard let config = config else { return }
 
-            LonaNode.run(
+            let rpcService = RPCService()
+
+            let stdinFileHandle = LonaNode.run(
```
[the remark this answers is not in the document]
dabbott: I think we should expose an RPC interface of some kind instead of a file handle. This would be OK on methods only used internally in this file, but for methods we call in other parts of LS a file handle should be an implementation detail (e.g. maybe the call actually goes over HTTP/websockets).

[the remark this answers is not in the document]
dabbott: Thinking out loud, maybe internally in the file we pass around a `Process` instead of file handles

mathieudutour:
All good questions:

1. 2. I only found libraries to parse and generate the JSON (like JSONRPCKit) which seems a bit useless to me. 

3. Even simples plugins would still need some info from Lona to run, right?

4. No that's the issue. I can't close stdin because I won't be able to reopen it. So I need to keep it open as long as the plugin is running and the plugin won't stop running until I close stdin _or_ if the plugin exit. That's why there is a `process.exit(0)` after the promise. That's really not ideal. Xi doesn't care about it since the process never stops.

studio/LonaStudio/Scripting/LonaNode.swift
```diff
@@ -9,6 +9,10 @@
 import Foundation
 import AppKit
 
+struct RunController {
```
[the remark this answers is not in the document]
mathieudutour: nothing, forgot to remove it

studio/LonaStudio/Plugins/LonaPlugins.swift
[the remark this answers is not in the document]
mathieudutour: So it doesn't prevent the node process from termination if the node process is not listening to stdin. It's only when the node process listens to it to it won't terminate (which is expected) and that's why there is an explicit `process.exit(0)` at the end of the plugin

studio/LonaStudio/Plugins/RPCService.swift
```diff
@@ -0,0 +1,197 @@
...
+    }
+    
+    static func ParseError(_ data: AnyObject? = nil) -> RPCError {
+        return RPCError(code: -32700, message: "Parse error", data: data)
+    }
+    
+    static func InvalidRequest(_ data: AnyObject? = nil) -> RPCError {
+        return RPCError(code: -32600, message: "Invalid Request", data: data)
+    }
+
```
lonabot: - :warning: Lines should not have trailing whitespace. (`trailing_whitespace`)

studio/LonaStudio/Plugins/RPCService.swift
```diff
@@ -0,0 +1,197 @@
...
+            let message = json["message"] as? String else { return nil }
+        self.code = code
+        self.message = message
+        self.data = json["data"]
+    }
+    
+    static func ParseError(_ data: AnyObject? = nil) -> RPCError {
+        return RPCError(code: -32700, message: "Parse error", data: data)
+    }
+
```
lonabot: - :warning: Lines should not have trailing whitespace. (`trailing_whitespace`)

studio/LonaStudio/Plugins/RPCService.swift
```diff
@@ -0,0 +1,197 @@
...
+    init?(json: [String: AnyObject]) {
+        guard let code = json["code"] as? Int,
+            let message = json["message"] as? String else { return nil }
+        self.code = code
+        self.message = message
+        self.data = json["data"]
+    }
+
```
lonabot: - :warning: Lines should not have trailing whitespace. (`trailing_whitespace`)

studio/LonaStudio/Plugins/RPCService.swift
```diff
@@ -0,0 +1,198 @@
...
+    static func InvalidRequest(_ data: AnyObject? = nil) -> RPCError {
+        return RPCError(code: -32600, message: "Invalid Request", data: data)
+    }
+
+    static func MethodNotFound(_ data: AnyObject? = nil) -> RPCError {
+        return RPCError(code: -32601, message: "Method not found", data: data)
+    }
+
```
lonabot: - :warning: Lines should not have trailing whitespace. (`trailing_whitespace`)

dabbott:
1-2. sounds good, let's just use our own stuff for now.

3. Currently some node "plugins" (just processes really) use info passed via argv/stdin. I think one-shot plugins are convenient and a simpler mental model when working in LS. So even if we instead pass info to them via RPC instead of argv at some point (or maybe read argv/stdin as JSON RPC), it's nice to have a convenience method to invoke them from LS as if they were functions, without the hassle of handling processes or pipes.

4. Ah, true. I guess we'll have to rely on plugins exiting themselves then. I could imagine in the future we could have a `health` JSON RPC call for long-lived processes and ping them to see if they're still doing OK -- if no response is received, we kill the process.

mathieudutour:
3. You can still pass arguments to the process like before, and if the plugin isn't listening to stdin (eg, doesn't use rpc), it will exit as expected. The API of PluginFile hasn't changed, it's still just `run(onSuccess: (String) -> Void)` so I'm not sure what the convenience method would look like.

4. I guess we could have a `health` check but I'm not sure i'd be useful to find plugins that have forgotten to call `process.exit()`: it'd be mostly for plugins with their event loop stuck

dabbott:
3. I'm mainly seeing the `_ = LonaNode.run`s in multiple places and wondering if we can get rid of the unused return value. There isn't a case where we actually do something with it yet.

4. yeah, maybe. I don't think we'll run into this for a long time, so we can think through it then.

mathieudutour:
Ah so a convenience method for `LonaNode` and not for `PluginFile`, got it

PR #309, opened 2019-02-14
fix crash on TextStyles and a few changes for plugins

mathieudutour:
## What

- fix closing the filehandle before writing the input
- add support for command field in plugin
- add new line character after JSON so that command can be separated
- add API request for compiler path

cc: @dabbott

### DIFF
--- a/studio/LonaStudio/Plugins/RPCService.swift
+++ b/studio/LonaStudio/Plugins/RPCService.swift
@@ -88,13 +88,21 @@ class RPCService {
     private func sendJson(_ json: Any) {
         do {
             let data = try JSONSerialization.data(withJSONObject: json, options: [])
-            sendData?(data)
+
+            // add a new line character so that the client can receive
+            // multiple answers at the same time
+            let mutdata = NSMutableData()
+            mutdata.append(data)
+            let nl = [0x0a as UInt8]
+            mutdata.append(nl, length: 1)
+
+            sendData?(mutdata as Data)
         } catch _ {
             print("error serializing to json")
         }
     }
 
-    private func sendResult(id: Int, result: Any) {
+    private func sendResult(id: Int, result: Any?) {
         let json = ["id": id, "result": result]
         sendJson(json)
     }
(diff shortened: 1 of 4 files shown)


studio/LonaStudio/Plugins/RPCService.swift
```diff
@@ -88,13 +88,21 @@ class RPCService {
     private func sendJson(_ json: Any) {
         do {
             let data = try JSONSerialization.data(withJSONObject: json, options: [])
-            sendData?(data)
+
+            // add a new line character so that the client can receive
+            // multiple answers at the same time
```
dabbott: When is this used? I don't totally know the flow of the RPC stuff yet... is this for batching calls?

mathieudutour: yeah, on the plugins I do `Promise.all([rpc.request('foo'), rpc.request('bar')])` and both answer arrives at the same time on stdin so I need a `\n` to separate them

dabbott: makes sense. that seems like it'll work, although I wonder if we should use an array instead and support this "batch" API https://www.jsonrpc.org/specification#batch? seems like a slightly simpler way to do the same thing

mathieudutour: We'd need to `\n` anyway because we don't get to control what are the batch sizes on stdin that node will read. But we could implement the `batch` API as well

dabbott: if we followed the `batch` format and sent an array, wouldn't studio receive a single request and send a single response? 

I think you're probably right that either way we need to add `\n`. Or use a streaming JSON parser, but probably not worth the effort.

mathieudutour: > if we followed the `batch` format and sent an array, wouldn't studio receive a single request and send a single response?

Yes but I think we should consider plugins as 3rd party black boxes so we can't guarantee that they won't send multiple individual requests

PR #313, opened 2019-02-18
Fix the component and types compiler commands

dabbott:
## What

The `component` and `types` commands stopped working after the compiler changes that introduced the `Config` object. This fixes them.

cc @mathieudutour


mathieudutour:
Can you publish a new version as well?

Issue #264, opened 2018-12-10
Add comment to the top of every generated file


dabbott:
## Summary

Add a comment containing the compiler's version number to the top of every file. This is useful to prevent people from unknowingly modifying generated code and having their changes wiped away. It's also helpful to see the compiler's version for debugging purposes.

We should add the comment:

```
// Generated by Lona Compiler 0.1.1
```

where the version number comes from the package.json. If it ends up looking nicer, we should add an extra newline after the comment.

We should do this for every Swift and JS file: components, colors, text styles, shadows, custom types, and static files copied from the `/static` dir. No need to handle Android XML files for now.

## Proposed Implementation

All file writing happens from `main.re`, so most or all of the changes will be in that file. We should prepend the comment string to the top of every relevant file before writing it. We shouldn't go through the AST printer for this, since that will complicate things and isn't necessary.

## Future Improvements

We should add a CLI parameter for modifying the comment, e.g. using a text file and prepending a project's LICENSE to every generated file.

PR #318, opened 2019-02-20
add header comment annotation

iammerrick:
## What

Fixes https://github.com/airbnb/Lona/issues/264

Still missing adding the annotation to files copied from `/static`

## Why

I did this change in order to familiarize myself with some of Lerna's implementation, it was tagged as a starter issue. :-)

## Testing Plan

- [ ] Added it to each invocation of generated file writing.

## Feedback

I'm looking for feedback on:

* Which files should I not have added it to 
* Which files should I have added it to
* Reason implementation

I'd love a general discussion around the `.component` language design. :-) 

cc: @dabbott, @ryngonzalez


dabbott:
Nice! Looks good to me. These are the right files to commit. Thanks for the contribution 😄

I'm always happy to chat about the `.component` design. I'll add you to the Lona slack.

dabbott:
Do you want to add the same annotation message to static files as part of this PR? Or should I go ahead and merge?

CI currently only tests Swift stuff, and not these files, so we don't need to wait for CI.

iammerrick:
Awesome! I'm looking forward to talking about it on slack! I'd love to merge this now and do static files as a separate PR. :-)

PR #321, opened 2019-03-02
JSON token file search + config command

dabbott:
## What

The tokens files (colors, textStyles, shadows, and gradients) may now be in workspace subdirectories. We also allow `*.colors.json`, etc. So `core.colors.json` is valid. Currently we only allow one of each kind of file, and it's a compiler error if there are 0 or 2 of any of them. We don't traverse into `node_modules`. The new project template places these files into a `foundation` directory. The `material-design` example project has been updated to follow the same pattern.

This also introduces a compiler command, `config`, that will print out the config used by the compiler internally. E.g:

```json
$ node src/main.bs.js config ../../examples/material-design | json_pp
{
   "paths" : {
      "colors" : "../../examples/material-design/foundation/colors.json",
      "shadows" : "../../examples/material-design/foundation/shadows.json",
      "textStyles" : "../../examples/material-design/foundation/textStyles.json",
      "components" : [
         "../../examples/material-design/components/elements/Card.component",
         "../../examples/material-design/components/elements/ListItem.component",
         "../../examples/material-design/components/screens/Team.component"
      ],
      "workspace" : "../../examples/material-design"
   },
   "version" : "0.5.2"
}
```

@mathieudutour


compiler/core/src/core/config.re
```diff
@@ -40,22 +40,46 @@ module Workspace = {
       );
   };
 
+  let findPath = (workspacePath: string, fileSuffix: string) => {
+    let searchPath = Path.join([|workspacePath, "**/*?(.)" ++ fileSuffix|]);
+    let searchResults =
+      FileSearch.sync(
+        searchPath,
+        ~options={ignore: ["**/node_modules"]},
```
mathieudutour: another one we should ignore is `.git`.

What do you think about adding an `ignore` field in lona.json to specify those directory that should be ignored? Also ignoring them for the .component search

compiler/core/src/core/config.re
```diff
@@ -194,4 +227,26 @@ let load =
...
+  Json.(
+    Encode.object_([
+      ("version", Encode.string(compilerVersion)),
+      (
+        "paths",
+        Encode.object_([
+          ("workspace", Encode.string(config.workspacePath)),
+          ("colors", Encode.string(config.colorsFile.path)),
+          ("textStyles", Encode.string(config.textStylesFile.path)),
+          ("shadows", Encode.string(config.shadowsFile.path)),
```
mathieudutour: isn't there `gradients` as well?

compiler/core/src/core/config.re
dabbott: That sounds good.

compiler/core/src/core/config.re
dabbott: Yeah. We don't generate any code for it yet though. I could put the path in if we wanted.

mathieudutour: Ah got. Nevermind then

compiler/core/src/core/config.re
dabbott: Updated (though not tested exhaustively). LS doesn't read from lona.json yet, only the compiler.

dabbott:
I just published `lonac@0.5.3` with these changes

PR #328, opened 2019-03-07
Fix #325: Pressing "Return" key opens the selected project

### LINKED ISSUE #325: [Lona Studio] Pressing "Enter" should open a project on welcome screen
## Summary

Currently the welcome screen supports keyboard navigation through the list of recently opened projects, but pressing the "Enter" key doesn't open a project.

screen_shot_2019-03-05_at_9_09_01_am

We're copying the Xcode welcome window's behavior here: we want double-click to open a project (which it currently does) and "Enter" to open a project, but we _don't_ want single-click to open a project.

## Implementation

The project list component is in `LonaStudio/Workspace/RecentProjectsList.swift`. I think overriding the `keyDown` method and handling the return key (`NSCarriageReturnCharacter`) will be simplest, but I'm open to anything that works.

klymenkoo:
## What
This PR contains fix for #325: it adds support for opening selected project from welcome screen when "Enter" was pressed.

mar-06-2019 21-01-59

In order to implement this feature, I overrode `keyDown` function inside the extension of `RecentProjectsTableView`, where I basically check if the keyCode of the pressed code is equalled to carriage return key code (36).

cc: @dabbott, @ryngonzalez

### DIFF
--- a/studio/LonaStudio/Workspace/RecentProjectsList.swift
+++ b/studio/LonaStudio/Workspace/RecentProjectsList.swift
@@ -172,6 +172,13 @@ extension RecentProjectsTableView: NSTableViewDelegate {
 
         onOpenProject?(projects[tableView.selectedRow])
     }
+
+	override func keyDown(with event: NSEvent) {
+		let carriageReturnKeyCode = 36
+		if event.keyCode == carriageReturnKeyCode {
+			onOpenProject?(projects[tableView.selectedRow])
+		}
+	}
 }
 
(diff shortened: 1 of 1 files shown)


dabbott:
Great job, looks good! Later today or tomorrow I'll post some more issues if you're interested in picking up another.

dabbott:
Oh! I just realized (after merging) that this will crash if no rows are selected, since `tableView.selected` is `-1` if there's no selection. I just made a quick fix and added `tableView.selectedRow >= 0` to the condition.

PR #330, opened 2019-03-08
move sketch library generation to a different package

mathieudutour:
## What

- [x] move the sketch library generation from the plugin to another package
- [x] use the package in the plugin

cc: @dabbott


mathieudutour:
>  we could have the sketch plugin's package.json point to the directory directly for easier local development

For dev, you can `yarn link` in the `sketch-library` folder and then `yarn link @lona/workspace-to-sketch-library`. But I'm thinking we will soon want the plugin to be bundled in the app or something so having a link there will broke it 

> At some point (not in this PR) we should figure out a better place for that device list. Maybe it could be in a JSON file that LS reads from also or something.

Agree

dabbott:
Ah yes, forgot about the `link` thing. Ok, sounds good.

PR #331, opened 2019-03-08
add babel step after compiler

mathieudutour:
## What

- [x] use babel as a step after the compiler instead of relying on `@babel/register`

cc: @dabbott


compiler/sketch-library/lib/index.js
```diff
@@ -5,8 +5,16 @@ const { exec } = require('child_process')
 const renderDocument = require('./render-document')
 const modifySketchTemplate = require('./modify-sketch-template')
 
+// https://gist.github.com/branneman/8048520#gistcomment-1249909
+// Add node_modules to the path, so they're resolved even when loading modules
+// from our compilerOutput directory (which is outside the root of this project)
+process.env.NODE_PATH = path.join(__dirname, 'node_modules')
+process.env.NODE_PATH = path.join(process.cwd(), 'node_modules')
```
dabbott: What does adding `process.cwd()` do?

mathieudutour: getting the directory from which the script is executed. That's the node_modules we care about. Actually, it should be the only one we add for production, the __dirname is only useful when linking it during dev

PR #336, opened 2019-03-11
Convert svg files using JSC instead of node

dabbott:
## What

Invoking the compiler to convert svg files was extremely slow.

I've included a bundle of the [svg converter](https://unpkg.com/@lona/svg-model@1.0.3/dist/svg-model.umd.js) directly in Lona Studio and run it using JSC. This is almost instant. 

> In the future we could also consider downloading non-essential scripts from unpkg on demand if we don't want to ship them all with Lona Studio


dabbott:
Travis seems to be having some trouble, so merging as-is.

PR #341, opened 2019-03-13
Fixes #339: Display SVG files

### LINKED ISSUE #339: [Lona Studio] Display SVG files when selected within the file navigator
## Summary

Currently Lona Studio displays some image files, e.g. PNG files, when selected in the file navigator. However, it doesn't display SVG files, which it really should, since it handles them elsewhere correctly.

Here's what the image viewer looks like for PNG files:

Screen Shot 2019-03-12 at 11 08 36 AM

And here's what it _currently_ looks like for SVG files:

Screen Shot 2019-03-12 at 11 08 43 AM

We want to use the same viewer as the PNG for SVG also.

## Proposed Implementation

I think we want to add `svg` to the list of supported file types:

Screen Shot 2019-03-12 at 11 12 17 AM

Then when we read the data for the `ImageDocument`
https://github.com/airbnb/Lona/blob/0571d6aa7c7796073eece13aa6a43acdb15840c7/studio/LonaStudio/Documents/ImageDocument.swift#L45 we should render the SVG file to an image
https://github.com/airbnb/Lona/blob/0571d6aa7c7796073eece13aa6a43acdb15840c7/studio/LonaStudio/Models/SVG.swift#L416

We might need to choose an arbitrary size for the image. We could use the `viewBox` of the svg to attempt to determine an "intrinsic" size of some kind, but that isn't really reliable. SVGs are designed to scale to any size, so they don't really have an intrinsic size, although `viewBox` is sometimes used like that. Maybe play around with it and see what looks best?

> If you want to try using the viewBox, it's a little more effort, but you can access it by first generating an `SVG.Node` before rendering an image, e.g. here https://github.com/airbnb/Lona/blob/0571d6aa7c7796073eece13aa6a43acdb15840c7/studio/LonaStudio/Models/SVG.swift#L241

klymenkoo:
## What
This PR contains a fix for #339: it displays SVG images, when the file was selected in file navigator: <br />
Screen Shot 2019-03-12 at 10 11 23 PM

### DIFF
--- a/studio/LonaStudio/Documents/ImageDocument.swift
+++ b/studio/LonaStudio/Documents/ImageDocument.swift
@@ -42,11 +42,11 @@ class ImageDocument: NSDocument {
 
     override func read(from url: URL, ofType typeName: String) throws {
         let data = try Data(contentsOf: url, options: NSData.ReadingOptions())
-		if url.pathExtension == "svg" {
-			let sz = CGSize(width: 450, height: 450)
-			content = SVG.render(contentsOf: url, size: sz, resizingMode: .scaleAspectFit)
-		} else {
-			content = NSImage(data: data)
-		}
+        if url.pathExtension == "svg" {
+            let size = CGSize(width: 450, height: 450)
+            content = SVG.render(contentsOf: url, size: size, resizingMode: .scaleAspectFit)
+        } else {
+            content = NSImage(data: data)
+        }
     }
 }
--- a/studio/LonaStudio/Documents/ImageDocument.swift
+++ b/studio/LonaStudio/Documents/ImageDocument.swift
@@ -42,6 +42,11 @@ class ImageDocument: NSDocument {
 
     override func read(from url: URL, ofType typeName: String) throws {
         let data = try Data(contentsOf: url, options: NSData.ReadingOptions())
-        content = NSImage(data: data)
+		if url.pathExtension == "svg" {
+			let sz = CGSize(width: 450, height: 450)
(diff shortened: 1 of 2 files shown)


studio/LonaStudio/Documents/ImageDocument.swift
```diff
@@ -42,6 +42,11 @@ class ImageDocument: NSDocument {
 
     override func read(from url: URL, ofType typeName: String) throws {
         let data = try Data(contentsOf: url, options: NSData.ReadingOptions())
-        content = NSImage(data: data)
+		if url.pathExtension == "svg" {
+			let sz = CGSize(width: 450, height: 450)
```
dabbott: We tend not to abbreviate variables in our codebase, so let's call this `size` instead

studio/LonaStudio/Documents/ImageDocument.swift
```diff
@@ -42,6 +42,11 @@ class ImageDocument: NSDocument {
 
     override func read(from url: URL, ofType typeName: String) throws {
         let data = try Data(contentsOf: url, options: NSData.ReadingOptions())
-        content = NSImage(data: data)
+		if url.pathExtension == "svg" {
```
dabbott: It looks like your indentation setting is set to tabs? Can you change it to 4 spaces?

Screen Shot 2019-03-13 at 5 46 29 PM

dabbott:
Awesome, great work!

I dunno what's up with the tests yet, but it's unrelated to this.

PR #351, opened 2019-03-21
fix recursive propTypes

mathieudutour:
## What

Recursive propTypes needs to be wrapped in a function

cc: @dabbott

### DIFF
--- a/compiler/generate-docs/src/components/Layout/Header.js
+++ b/compiler/generate-docs/src/components/Layout/Header.js
@@ -85,15 +85,6 @@ const Header = ({ data, location, files }) => (
   </Wrapper>
 )
 
-const FilesPropTypes = PropTypes.objectOf(
-  PropTypes.shape({
-    path: PropTypes.string.isRequired,
-    title: PropTypes.string,
-    order: PropTypes.number,
-    children: FilesPropTypes, // eslint-disable-line no-use-before-define
-  })
-)
-
 Header.propTypes = {
   location: PropTypes.shape({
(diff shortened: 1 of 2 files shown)


ljharb:
PropTypes can definitely be recursive - you’d need to make the child one a function that wraps the outer one, though, I’ve done it many times.

mathieudutour:
Ah indeed, I didn't know. Thanks

compiler/generate-docs/src/components/Layout/Header.js
```diff
@@ -90,7 +90,7 @@ const FilesPropTypes = PropTypes.objectOf(
     path: PropTypes.string.isRequired,
     title: PropTypes.string,
     order: PropTypes.number,
-    children: FilesPropTypes, // eslint-disable-line no-use-before-define
+    children: () => FilesPropTypes,
```
ljharb:
```suggestion
    children: (...args) => FilesPropTypes(...args),
```

compiler/generate-docs/src/components/Layout/Sidebar.js
```diff
@@ -167,7 +167,7 @@ const FilesPropTypes = PropTypes.objectOf(
     path: PropTypes.string.isRequired,
     title: PropTypes.string,
     order: PropTypes.number,
-    children: FilesPropTypes, // eslint-disable-line no-use-before-define
+    children: () => FilesPropTypes,
```
ljharb:
```suggestion
    children: (...args) => FilesPropTypes(...args),
```

PR #354, opened 2019-03-23
Update generated docs text styles and spacing

dabbott:
## What

Misc updates to text styles and spacing.

It's always tricky jumping into CSS stuff, hopefully I didn't mess anything up!

localhost_8000_u_components_elements_Card (1)

@mathieudutour


dabbott:
That would be good, but since LS uses the built-in Swift JSON encoder, I'm not sure if there's an easy way to do that. The pretty printing is pretty much one-size-fits-all: https://developer.apple.com/documentation/foundation/jsonencoder/outputformatting

mathieudutour:
Ah that makes sense

PR #357, opened 2019-03-27
Introducing Cypress and Applitools

megsachdev:
Hello!
At times maintaining separate visual test cases becomes a tiring task so using your existing examples setup we tried coming up with something that can ease this process. So I chose to use Cypress as the browser test harness, which Applitools has an SDK for. I started off with writing tests for some of the components so it's sort of a "smoke test" but you can add more if you'd like.

Visual regression involves comparing the pixels visually like a human would, ideally ignoring differences that a human couldn't perceive e.g. 1 pixel changes, differences in how browsers render images, drop shadows, etc. to avoid false positives; but you can still do a strict 1:1 pixel match if that's your preference.

Applitools thankfully provides an SDK that made it trivial to write tests using Cypress. So other than the things I included what you'll need to do is create your own Applitools account and then set up your GitHub CI integration if you want it to run automatically on Pull Requests; optional, but probably a good idea.

To run Applitools:

Create Applitools account
Go to the menu in the top right corner and click the profile icon on the far right.
keysetup
Click on "My API Key" and copy your key
Go to your terminal and paste
`export APPLITOOLS_API_KEY=Your_API_Key_Here`
Go to your directory and run:
`yarn cypress:visual`
In your Applitools dashboard you should see the results of your tests
homepage

### DIFF
--- a/examples/LonaViewer/web/.gitignore
+++ b/examples/LonaViewer/web/.gitignore
@@ -23,3 +23,7 @@ src/generated
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+
+# cypress
+cypress/screenshots
+cypress/videos
--- a/examples/LonaViewer/web/cypress.json
+++ b/examples/LonaViewer/web/cypress.json
@@ -1,0 +1,3 @@
+{
+  "baseUrl": "http://localhost:3000/"
+}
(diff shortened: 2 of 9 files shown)


dabbott:
This looks awesome! I should have some time to check this out this weekend, but if not, first thing next week.

Are you working with Tracy and friends at https://www.thisdot.co/?

megsachdev:
Thank you.. yup I am working with Tracy..

megsachdev:
Yes you can go ahead and add tests for the components you want. This is basically useful for visual regression testing and is helpful in case anything breaks due to code change, it's efficient enough to catch the same.

PR #359, opened 2019-03-29
Fixes #355: Add component duplication support

### LINKED ISSUE #355: [Lona Studio] Add easy component duplication
## Summary

When creating a new component, it's common to duplicate an exist component as a starting point. There's an option for this in the app menu under `File > Duplicate`, but it would be more convenient/discoverable if it were a context menu item in the File Navigator:

Screen_Shot_2019-03-23_at_3_45_48_PM

Let's add a new `Duplicate As...` menu item when right-clicking components in the File Navigator. Upon clicking the menu item, we will:

1. Show a dialog prompting the user to choose the filename for the new component
2. Copy the selected file to the chosen filename, and open the new file as a document

## Proposed Implementation

1. We'll be adding a new menu item to the File Navigator here: https://github.com/airbnb/Lona/blob/9ee49dfbb9d3dce6309c90b93c51c6949cf4f214/studio/LonaStudio/Workspace/FileNavigator.swift#L188 

Note that: 
- just like for the `Delete` menu item, we don't want to show this menu item for the root path
- unlike for `Delete`, this `Duplicate As...` menu item should only be available for components, not directories, assets, `.json` files, or anything else.

2. Upon clicking the menu item, we should show a "save" dialog for choosing a filename. For an example of how to show a "save" a dialog, see: https://github.com/airbnb/Lona/blob/9ee49dfbb9d3dce6309c90b93c51c6949cf4f214/studio/LonaStudio/Workspace/WorkspaceViewController.swift#L659-L674

    If possible, we should prevent the user from choosing a file path outside of the root directory, or show an `Alert` if they do and prevent the operation.

3. Copy the selected file to the chosen destination. The easiest way to copy a file is probably `FileManager.default.copyItem`. 

4. Open the new file. The method we need to call is already passed into the File Navigator as `onAction`: https://github.com/airbnb/Lona/blob/9ee49dfbb9d3dce6309c90b93c51c6949cf4f214/studio/LonaStudio/Workspace/WorkspaceViewController.swift#L399

klymenkoo:
This PR contains a fix for #355.

It adds a new item "Duplicate as..." to the context menu. If the user selects this option, the app will prompt the user to type a name for this component and select location for it. Once the user saves the file, it's going to be opened in the Component Editor.

**Implementation**:
1. Check the extension of the selected file.
2. Add menu item "Duplicate As..."
3. onClick it shows save dialog.
4. Try to copy the item to the new path.
5. If successful, open a newly created file. Otherwise, it will show an alert with message "Couldn't copy component to..."

### DIFF
--- a/studio/LonaStudio/Workspace/FileNavigator.swift
+++ b/studio/LonaStudio/Workspace/FileNavigator.swift
@@ -185,6 +185,36 @@ class FileNavigator: NSBox {
                 menu.addItem(NSMenuItem.separator())
             }
 
+            if NSURL(fileURLWithPath: path).pathExtension == "component"{
+                menu.addItem(NSMenuItem(title: "Duplicate As...", onClick: {
+                    var saveURL: String
+                    let dialog = NSSavePanel()
+
+                    dialog.title                   = "Save .component file"
+                    dialog.showsResizeIndicator    = true
+                    dialog.showsHiddenFiles        = false
+                    dialog.canCreateDirectories    = true
+                    dialog.allowedFileTypes        = ["component"]
+
+                    // User canceled the save. Don't swap out the document.
+                    if dialog.runModal() != NSApplication.ModalResponse.OK {
+                        return
+                    }
+
+                    guard let url = dialog.url else { return }
+                    saveURL = url.path
+                    do {
+                        try FileManager.default.copyItem(atPath: path, toPath: saveURL)
+                        self.onAction?(saveURL)
+                    } catch {
+                        let alert = NSAlert()
+                        alert.messageText = "Couldn't copy componento to \(saveURL)"
+                        alert.addButton(withTitle: "OK")
+                        alert.runModal()
+                    }
+                }))
+            }
+
             menu.addItem(NSMenuItem(title: "Delete", onClick: {
                 let fileName = URL(fileURLWithPath: path).lastPathComponent
 
--- a/studio/LonaStudio/Workspace/FileNavigator.swift
+++ b/studio/LonaStudio/Workspace/FileNavigator.swift
@@ -185,9 +185,8 @@ class FileNavigator: NSBox {
                 menu.addItem(NSMenuItem.separator())
             }
 
-            if NSURL(fileURLWithPath: path).pathExtension == "component"{
+            if URL(fileURLWithPath: path).pathExtension == "component" {
                 menu.addItem(NSMenuItem(title: "Duplicate As...", onClick: {
-                    var saveURL: String
                     let dialog = NSSavePanel()
 
                     dialog.title                   = "Save .component file"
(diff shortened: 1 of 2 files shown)


studio/LonaStudio/Workspace/FileNavigator.swift
```diff
@@ -185,6 +185,36 @@ class FileNavigator: NSBox {
                 menu.addItem(NSMenuItem.separator())
             }
 
+            if NSURL(fileURLWithPath: path).pathExtension == "component"{
```
dabbott: Nit: there should be a space after "component". I'm not sure why SwiftLint doesn't warn on that.

studio/LonaStudio/Workspace/FileNavigator.swift
```diff
@@ -185,6 +185,36 @@ class FileNavigator: NSBox {
...
+                    // User canceled the save. Don't swap out the document.
+                    if dialog.runModal() != NSApplication.ModalResponse.OK {
+                        return
+                    }
+
+                    guard let url = dialog.url else { return }
+                    saveURL = url.path
+                    do {
+                        try FileManager.default.copyItem(atPath: path, toPath: saveURL)
+                        self.onAction?(saveURL)
```
dabbott: If we call `self.onAction?(saveURL)` in the `do ... catch` block, it will swallow errors thrown when opening the file, which probably isn't what we want.

Let's move this call to after the `do ... catch` block. We can add a `return` to the end of the `catch`.

studio/LonaStudio/Workspace/FileNavigator.swift
```diff
@@ -185,6 +185,36 @@ class FileNavigator: NSBox {
...
+                    dialog.canCreateDirectories    = true
+                    dialog.allowedFileTypes        = ["component"]
+
+                    // User canceled the save. Don't swap out the document.
+                    if dialog.runModal() != NSApplication.ModalResponse.OK {
+                        return
+                    }
+
+                    guard let url = dialog.url else { return }
+                    saveURL = url.path
```
dabbott: I would either: 
- remove the `saveURL` variable and call `url.path` directly when we need it (simpler, 1 fewer var) 
- move the variable declaration down to the same line (`let saveURL = url.path`) and rename `saveURL` to `savePath`, since it's not actually a `URL` anymore

studio/LonaStudio/Workspace/FileNavigator.swift
dabbott: Let's use `URL` instead of `NSURL`. I think they're aliases, but we use `URL` everywhere else.

studio/LonaStudio/Workspace/FileNavigator.swift
```diff
@@ -185,6 +185,36 @@ class FileNavigator: NSBox {
                 menu.addItem(NSMenuItem.separator())
             }
 
+            if NSURL(fileURLWithPath: path).pathExtension == "component"{
+                menu.addItem(NSMenuItem(title: "Duplicate As...", onClick: {
+                    var saveURL: String
+                    let dialog = NSSavePanel()
```
dabbott: I just pulled down this branch and tried it out. I think the most common use case for this feature is to duplicate a component into the same directory. Let's set `dialog.directoryURL` to make that easier. We can do this with `dialog.directoryURL = URL(fileURLWithPath: path).deletingLastPathComponent()`. I didn't think of this before, but I think it'll help a lot.

PR #360, opened 2019-03-30
Update LS to swift 5

mathieudutour:
## What

- [x] fix some warning in Xcode 10.2
- [x] update to Swift 5 (stable ABI FTW!)

cc: @dabbott


dabbott:
Nice! Thanks for doing this. I'm going to try to get the travis build to pass before we merge. I just updated Xcode to 10.2, which I think will solve it.

PR #361, opened 2019-03-30
no more LineCapStyle in swift 5

mathieudutour:
## What

Swift 5 removes `LineCapStyle` on Appkit (see #360).
I might be worth removing the switch completly?

cc: @dabbott


dabbott:
I think we might need to support Swift 4 still. @outdooricon which version of Swift are y'all using? We may need to add a CLI param for Swift version.

The other Swift 4 vs 5 difference that's probably relevant for us is `NSAttributedStringKey` => `NSAttributedString.Key`. Those are in the static files `TextStyle.uikit.swift` and `TextStyle.appkit.swift`. We could either create separate static files, e.g. `TextStyle.appkit.swift5`, or do a big string literal template like https://github.com/airbnb/Lona/blob/9ee49dfbb9d3dce6309c90b93c51c6949cf4f214/compiler/core/src/swift/swiftCollectionView.re#L1

We could do that in a separate PR, but let's first figure out if we still need Swift 4 or not.

dabbott:
Yeah, it looks like we still need Swift 4 for now. As far as I can tell it's just the lineCapStyle and NSAttributedStringKey. I'd probably go with the `TextStyle.appkit.swift5` approach since it seems simpler. I could do that or you, either way. Let's just lump it all in this PR.

dabbott:
FYI I'm going to do this now

mathieudutour:
Thanks for finishing that 🙏

PR #363, opened 2019-04-01
update swiftlint config

mathieudutour:
## What

- [x] update the swiftlint config to match https://github.com/airbnb/swift


mathieudutour:
I was surprised as well!

PR #364, opened 2019-04-01
consistent linting defined at the root

mathieudutour:
## What

- [x] get rid of all the eslint/prettier config and define one at the root


README.md
```diff
@@ -58,12 +58,11 @@ The Lona Compiler converts .component files to UI code for various targets.
 Currently, these targets are supported:
 
 - iOS / macOS (Swift)
-- React DOM / React Native (JavaScript)
+- React DOM / React Native / React Sketchapp (JavaScript)
```
dabbott: :+1:

mathieudutour:
I do and it does when you open the root folder (that's how I fixed most of the lint errors). Not sure about opening a specific sub-folder tho

dabbott:
:+1: that should be good then. The only subfolder I tend to open is the compiler, and that uses `refmt` anyway

mathieudutour:
I also added a pre-commit hook to lint the swift/js/jsx files and run prettier. I could add `refmt` to the mix?

dabbott:
Let's leave off `refmt` for now, but something to do in the future. I still have some trouble locking down its version correctly, and don't want to muck around with that right now.

The pre-commit hook is good. I didn't really notice in my first look through, but we added 4 new files to the root dir... can we try to consolidate? Having a lot of files in the root can be kind of intimidating for beginners/designers I think. Both of the `eslintConfig`/`eslintIgnore` should be able to go in the `package.json`. There's probably no way to avoid `package.json` and `yarn.lock` in the root, given that that's the behavior we want, so that's fine.

Issue #356, opened 2019-03-23
[Lona Studio] Implement bottom pane hide/show


dabbott:
## Summary

This button _should_ hide/show the bottom section of the UI, but currently it does nothing!

![Screen_Shot_2019-03-23_at_4_16_20_PM](https://user-images.githubusercontent.com/1198882/54872785-32d11e00-4d87-11e9-9075-6e0fc1977e98.png)

Let's fix that. Some new users have been pretty confused by why the bottom hasn't been responding...

## Proposed Implementation

When the `activePanes` change... https://github.com/airbnb/Lona/blob/9ee49dfbb9d3dce6309c90b93c51c6949cf4f214/studio/LonaStudio/Workspace/WorkspaceViewController.swift#L84

We currently ignore the bottom pane: https://github.com/airbnb/Lona/blob/9ee49dfbb9d3dce6309c90b93c51c6949cf4f214/studio/LonaStudio/Workspace/WorkspaceViewController.swift#L301

Instead, we need to hide/show the `bottomItem` in the `ComponentEditorViewController`: https://github.com/airbnb/Lona/blob/9ee49dfbb9d3dce6309c90b93c51c6949cf4f214/studio/LonaStudio/Workspace/ComponentEditorViewController.swift#L213-L216

The easiest way to do this is probably to expose the `bottomItem` as a _public or internal_ variable of the `ComponentEditorViewController`, but this isn't a great approach, since then we're exposing implementation details to a parent view controller. If we make `bottomItem` a member variable, it should be _private_.

Instead, let's add a variable `var utilitiesViewVisible: Bool` (with a `didSet` block) to `ComponentEditorViewController`, which we can then update from the `WorkspaceViewController`. The `WorkspaceViewController` contains an instance of `ComponentEditorViewController` we can use: https://github.com/airbnb/Lona/blob/9ee49dfbb9d3dce6309c90b93c51c6949cf4f214/studio/LonaStudio/Workspace/WorkspaceViewController.swift#L127

We'll also need to add/copy something similar to `setVisibility` into `ComponentEditorViewController` to do the actual hiding/showing.
https://github.com/airbnb/Lona/blob/9ee49dfbb9d3dce6309c90b93c51c6949cf4f214/studio/LonaStudio/Workspace/WorkspaceViewController.swift#L305-L315

[...]

PR #374, opened 2019-04-11
Hide/show bottom panel

klymenkoo:
I would like to share my progress on #356. I've tried to implement the suggested changes from the issue description. This is what I have so far:
If the bottom panel is in its initial state (the app has been just started, and the user selected component), the hide/show doesn't work:
hide-show-1

If the user resizes the bottom panel, the hide/show button works fine:
hide-show-fix

### DIFF
--- a/studio/LonaStudio/Workspace/WorkspaceViewController.swift
+++ b/studio/LonaStudio/Workspace/WorkspaceViewController.swift
@@ -85,7 +85,7 @@ class WorkspaceViewController: NSSplitViewController {
         get {
             return WorkspacePane.all.filter {
                 if $0 == .bottom {
-                    return false
+                    return !(componentEditorViewController.utilitiesViewVisible)
                 } else {
                     return !(splitViewItem(for: $0)?.isCollapsed ?? true)
                 }
--- a/studio/LonaStudio/Workspace/WorkspaceViewController.swift
+++ b/studio/LonaStudio/Workspace/WorkspaceViewController.swift
@@ -84,7 +84,11 @@ class WorkspaceViewController: NSSplitViewController {
     public var activePanes: [WorkspacePane] {
         get {
             return WorkspacePane.all.filter {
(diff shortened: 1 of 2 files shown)


studio/LonaStudio/Workspace/WorkspaceViewController.swift
```diff
@@ -84,7 +84,11 @@ class WorkspaceViewController: NSSplitViewController {
     public var activePanes: [WorkspacePane] {
         get {
             return WorkspacePane.all.filter {
-                return !(splitViewItem(for: $0)?.isCollapsed ?? true)
+                if $0 == .bottom {
+                    return false
```
klymenkoo: Since I don't have a public bottomItem, what should I do here?
Should I add a public getter for `isCollapsed` property of bottomItem in the `componentEditorViewController`, and use it here?

dabbott: Would it work to use `utilitiesViewVisible` for this, with a `get` and `set` block instead of a `didSet`? E.g.

```swift
public var utilitiesViewVisible: Bool {
  get { return bottomItem.isCollapsed }
  set { setBottomItemVisibility(to: newValue) }
}
```

Then we'd only need one public variable for collapsed state

dabbott:
I think we want to increase the minimum thickness of the bottom pane so that it can't be in that weird state where it has 0 height but it's not "collapsed": https://github.com/airbnb/Lona/blob/31e6ffdc66703efd2198cd84688904cc02daa8ce/studio/LonaStudio/Workspace/ComponentEditorViewController.swift#L215 I'm not sure what the best value would be... maybe try 200? Or play around with it and see what seems like a good default

klymenkoo:
@dabbott 200 worked perfectly 👍

klymenkoo:
I've implemented your suggestions, and it seems to work fine now 🙌

klymenkoo:
@dabbott I believe so.

PR #376, opened 2019-04-12
rework events to have a paylaod and implement preview cache

mathieudutour:
## What

- [x] rework events so that they can be triggered with a payload. I don't really like the implementation but I'm not sure there is a way to have an `eventType` and a callback type that depends on the `eventType`.
- [x] implement component preview cache

cc: @dabbott


dabbott:
Ok, I took a look, and I agree that the implementation doesn't seem ideal. I'm not sure if there's a better way, but I'd probably need to pull it down and fiddle around with the types for a while to come up with alternatives.

I think this track of work is important (faster loading, caching, and reloading of images, components, etc), but I want to focus on the file formats and core editing experience this week, so let's revisit this next week

PR #379, opened 2019-04-13
[RFC] new colors format

mathieudutour:
[Rendered view](https://github.com/airbnb/Lona/blob/7dd7986357a4aa4d24ba38cf944c1a992ea1d04d/docs/file-formats/colors.md)

cc: @dabbott


docs/file-formats/colors.md
```diff
@@ -2,26 +2,123 @@
...
+
+The value of a color is represented by a JSON object. There are multiple ways to represent it, hence there are multiple set of properties available to represent the color.
+
+- `hex`: `{ "hex": "#00FF00" }`
+- `rgb`: `{ "r": 0, "g": 1, "b": 0 }`
+- `hsl`: `{ "h": 120, "s": 1, "l": 0.5 }`
+- `hsv`: `{ "h": 120, "s": 1, "v": 1 }`
+
+Only one set of properties must be specified for a single value.
+
```
mathieudutour: I'm also wondering if we should start to include an `alpha` property

docs/file-formats/colors.md
```diff
@@ -2,26 +2,123 @@
...
+> The reason why it's an array instead of an object with the id as keys is that we want to order colors, and it's really not recommended to rely on the order of the keys of a JSON object.
+
+### Color Value
+
+#### Value
+
+The value of a color is represented by a JSON object. There are multiple ways to represent it, hence there are multiple set of properties available to represent the color.
+
+- `hex`: `{ "hex": "#00FF00" }`
+- `rgb`: `{ "r": 0, "g": 1, "b": 0 }`
```
mathieudutour: wondering if we should use `r`, `g`, `b` or `red`, `green`, `blue`. Might not matter too much for `rgb` but could be clearer for `hsl`

docs/file-formats/colors.md
```diff
@@ -2,26 +2,123 @@
...
+    "id": "baseColor",
+    "name": "Base Color",
+    "value": { "hex": "#00FF00" }
+  }
+  {
+    "id": "desaturatedBaseColor",
+    "name": "Desaturated Base Color",
+    "value": {
+      "extends": "baseColor",
+      "s": 0.5
```
dabbott: Ooh 👍 for color adjustments

I might even say, let's take it a step further and allow any Lona expression (even if we only implement the very basics to begin with). That'd make it pretty easy to generalize to aliases, color adjustments, or more advanced things potentially.

```swift
struct Color {
  indirect enum Value {
    case color(String)
    case group([Color])
    case ref(LonaExpression)
  }
}
```

Where `LonaExpression` can be things like:

```js
// An identifier, for aliasing
"case": "ref",
"value": "baseColor"

// A member expression, for aliasing (could be dot-separated instead, if every member is an identifier)
"case": "ref",
"value": ["blues", "100"]

// A function call expression (idk exactly the format, but something like)
"case": "ref",
"value": {
  "type": "call",
  "id": "saturate",
  "arguments": ["baseColor", "20"]
}
```

We could optionally have a more concise format for simple/common adjustments. 

```json
"case": "saturate",
"value": {
  "base": "baseColor",
  "s": "20"
}
```

For complicated expressions, e.g. nested adjustment functions, people would definitely want to make them in LS, so I think it's OK if the JSON gets complicated too

In the future, it'd be cool if we could even generate a whole palette or group of colors dynamically. Like, palette generations plugins

docs/file-formats/colors.md
mathieudutour: The issue with CSS is that it's pretty limited: you only have 255*255*255 values, there is no color profile support, etc. I wouldn't be surprised if there is an extension to CSS sooner than later around wide gamut screen support and that kind of things.

The compiler can easily translate a more complete color value description down to CSS if needed.

docs/file-formats/colors.md
```diff
@@ -2,26 +2,123 @@
...
 
-At the top level, the file is an object containing a `"colors"` array.
+At the top level, the file is an object containing:
+
+- a `theme` array
+- a `colors` array
+
+### `theme`
+
+The `theme` array is optional. The default theme is the first item of the array, if any.
```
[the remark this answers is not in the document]
mathieudutour: Yes I was thinking about having a separate `themes.json` file but couldn't think of any use cases where theme applies to anything other than colors (except maybe syntax highlight theme where you might want to have italic font variants but that's really an edge case indeed)

docs/file-formats/colors.md
```diff
@@ -2,26 +2,123 @@
...
+
+Each theme has the following attributes:
+
+| Attribute  | Type     | Required | Description                                                                                                                                                                                                                                                                                                            |
+| ---------- | -------- | -------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `id`       | `string` | Yes      | The unique `id` of the theme, which will be used internally by Lona Studio and in the generated code. This should be _code-friendly_: it must not contain spaces or unusual characters, since it will be used as a variable name. It should also be strictly longer than 1 character and cannot be equal to `extends`. |
```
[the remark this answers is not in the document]
mathieudutour: Because it would conflict with how you declare the color value:
```js
{
  themes: [{ id: 'r' }],
  colors: [{
    id: 'color',
    value: { r: 0, g: 0, b: 0 } } // does `r` reference the theme or `red`?
  ]
}
```

docs/file-formats/colors.md
```diff
@@ -2,26 +2,123 @@
...
+| `name`     | `string` | Yes      | The human-readable name of the theme. This will be displayed in the Lona Studio UI. This will not appear in the generated code.                                                                                                                                                                                        |
+| `comment`  | `string` | No       | An optional description of the theme, explaining contextual information, such as how it should be used. This may be displayed in the Lona Studio UI. This will not appear in the generated code.                                                                                                                       |
+| `metadata` | `object` | No       | Additional data used by the generated code to match platform specific features (for example `metadata.isMacOSDarkTheme` indicates that the theme should be used when the macOS system theme is `dark`).                                                                                                                |
```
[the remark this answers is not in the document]
mathieudutour: We do want a layer of indirection yes but the default behavior of macOS is to indeed control which theme is used.

I was thinking that the boilerplate code could be auto-generated when `metadata.isMacOSDarkTheme` is set but didn't really think it through so it might not make sense

docs/file-formats/colors.md
```diff
@@ -2,26 +2,123 @@
...
+| `id`       | `string`  | Yes                                | The unique `id` of the color, which will be used internally by Lona Studio and in the generated code. This should be _code-friendly_: it must not contain spaces or unusual characters, since it will be used as a variable name. |
+| `name`     | `string`  | Yes                                | The human-readable name of the color. This will be displayed in the Lona Studio UI. This will not appear in the generated code.                                                                                                   |
+| `value`    | `object`  | Required if there is no `variants` | An object representing the color. See [Color Value](#color-value) for more information about the properties of the object.                                                                                                        |
+| `variants` | `Color[]` | Required if there is no `value`    | An array representing the variants of the color.                                                                                                                                                                                  |
```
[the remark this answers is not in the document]
mathieudutour: That makes sense yes

docs/file-formats/colors.md
```diff
@@ -2,26 +2,123 @@
...
+  "theme": [
+    { "id": "light", "name": "Light Theme" },
+    { "id": "dark", "name": "Dark Theme" }
+  ],
+  "colors": [
+    {
+      "id": "textColor",
+      "name": "Text Color",
+      "value": {
+        "hex": "#000000",
```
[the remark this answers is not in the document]
mathieudutour: I'm trying to think of a case where you would want to define the value for just one theme but not others and can't so it makes sense to add that constraint :+1:

docs/file-formats/colors.md
mathieudutour: I like that! Where would the functions be declared? Are they pre-defined?

docs/file-formats/colors.md
dabbott: That makes sense, I didn't know about the new CSS color spec. In that case, howabout we support the proposed syntax? https://www.w3.org/TR/css-color-4/#changes-from-3

We don't have to support it all at once, but can start with the things that seem most important/relevant for us

docs/file-formats/colors.md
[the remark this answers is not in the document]
dabbott: l like the idea, we can think it through a bit more when we get to the code-gen for this stuff

docs/file-formats/colors.md
dabbott: I think pre-defined to start, and then maybe at some point we can make them extendable

mathieudutour:
Sorry for the delay. I updated the spec to match what we said

docs/file-formats/json-schemas/colors.json
```diff
@@ -0,0 +1,151 @@
+{
```
mathieudutour: I added a JSON schema which is really useful when looking at a colors file with VS Code. It shows hints on hover and validate the JSON against the schema automatically.

I think it also wouldn't be hard to generate the markdown spec from it (the other way around is probably harder)

ChristophLabacher:
Really like the additional features! References are something that I was looking forward to as well.

Looking at your example, I wonder why a color can either be themed or a ref. I’d imagine that often times when switching themes you don’t want to change your visual color values, but switch your semantic color assignments to a different visual color (which you reference). For example:

```json
{
  "theme": [
    { "id": "light", "name": "Light Theme" },
    { "id": "dark", "name": "Dark Theme" }
  ],
  "colors": [
    {
      "id": "textColor",
      "name": "Text Color",
      "case": "themed",
      "value": {
        "light": ["Black"],
        "dark": ["White"]
      }
    },
    {
      "id": "white",
      "name": "White",
      "case": "color",
      "value": "#FFFFFF"
    },
    {
      "id": "black",
      "name": "Black",
      "case": "color",
      "value": "#222222"
    }
  ]
}
```

(In reality you would probably have a group of visual colors and a group of semantic color assignments referencing those)

Maybe there is a different way of marking a value as a reference?

Something else that could potentially be valuable is custom metadata – to enable people to build their own toolchain around the Lona spec it might be good to have a key other than comment, under which people can add whatever information they want. A few years ago I was part of the team working on Open Color Tools, where we defined a format for color specifications too ([Introduction to OCO](https://github.com/opencolor-tools/opencolor-spec/blob/master/a_simple_introduction.md)). The ability to have arbitrary metadata assigned to groups and colors was very valuable for the whole ecosystem of tools surrounding it. Such a key might be helpful for all the specifications, not just color.

mathieudutour:
> I wonder why a color can either be themed or a ref

You mean a value of the theme color can either be a string or a ref?

I guess that can make sense depending on how you define your colors

ChristophLabacher:
> You mean a value of the theme color can either be a string or a ref?

Exactly! In my mind it seems very likely that if there is no theme a semantic color (like "Text Color") would be a ref to a visual color (like "Black"). I’d argue that that makes equal sense when there are themes.

docs/file-formats/colors.md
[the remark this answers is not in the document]
ChristophLabacher: For Type Styles breakpoints might play a similar role as themes for colors.

You could have one Text Style "large" with font size 64px and one "small" with font size 32px. Depending on the breakpoint (e.g. desktop or mobile) your Text Style "headline" could extend one or the other.

PR #385, opened 2019-05-12
generate doc and schema from yml

mathieudutour:
## What

I managed to generate the markdown file from the json schema. It found some inconsistencies and fixed them but as you can see, the diffs are minimal.

But JSON has a really poor authoring experience for all the multilines strings. So I decided to use YAML instead which is great.

And I added a couple of scripts to generate the JSON schema and the markdown documentation from the YAML file

cc: @dabbott


dabbott:
Nice! This is really cool. The YAML looks like a big improvement over authoring in JSON.

I didn't really look/think through the details of some of the more complex types of colors, e.g. `call`, but I think we can tweak that if needed whenever we get to implementing

mathieudutour:
> we can tweak that if needed whenever we get to implementing

Indeed, the markdown generation is very rough right now, probably doesn't handle anything outside the current file but we can improve it over time, as we need it

PR #388, opened 2019-05-21
Use Arg module for better CLI usage/reporting

dabbott:
## What

This allows running commands like:

`lonac workspace --target swift --workspace examples/test --output examples/generated/test/swift`

This also adds a `help` menu and better invalid option reporting.

This update should be 99% backwards compatible, e.g. positional arguments will still work, but they won't be recommended in the help menu. The only things that are different are `--debugConstraints` and `--generateCollectionView`, which no longer require a value, since they're booleans.

cc @mathieudutour this shouldn't break the docs/sketch plugin, but if a command stops working I wouldn't be surprised if it was due to this.

The help menu:

```
The Lona compiler generates code from a Lona workspace (a directory with a lona.json manifest file).

The following commands are available:

  workspace
  component
  colors
  textStyles
  shadows
  types
  config
  version

Type `lonac [command]` to see which options are required for that command. The following options are available:

  --target {js|swift|xml}
                               Output language
  --workspace [path]           Workspace directory path
  --output [path]              Output directory path
  --input [path]               Input file path (for commands that take an input file)
  --filterComponents [pattern] Only generate components matching this string (converted to a regular expression)
  --framework {uikit|appkit|reactdom|reactnative|reactsketchapp}
                               UI framework
  --swiftVersion {4|5}
                               Swift version
  --debugConstraints           Generate constraint IDs for easier debugging
  --generateCollectionView     Generate experimental UICollectionView subclass that renders LonaModels as cells
  --typePrefix [prefix]        Generated type prefix
  --styleFramework {none|styledcomponents}
                               Style framework
  --styledComponentsVersion {3|latest}
                               styled-components library version
  --minimumIeSupport {none|11}
                               Minimum version of IE supported
  -help                        Display this list of options
  --help                       Display this list of options
```

### DIFF
--- a/compiler/core/package.json
+++ b/compiler/core/package.json
@@ -21,7 +21,7 @@
     "snapshotJsNative:watch": "nodemon --exec \"yarn snapshotJsNative\"",
     "snapshotSketch": "node src/main.bs.js workspace js ../../examples/test ../../examples/generated/test/sketchJs --framework=reactsketchapp",
     "snapshotSketch:watch": "nodemon --exec \"yarn snapshotSketch\"",
-    "snapshotSwift": "node src/main.bs.js workspace swift ../../examples/test ../../examples/generated/test/swift --generateCollectionView=1",
+    "snapshotSwift": "node src/main.bs.js workspace swift ../../examples/test ../../examples/generated/test/swift --generateCollectionView",
     "snapshotSwift:watch": "nodemon --exec \"yarn snapshotSwift\"",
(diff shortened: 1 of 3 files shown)


mathieudutour:
when looking at `git` for example, it doesn't list the options when showing the help for `git`. It only shows them when looking at the help for a specific command.

If the number of options starts to grow, we might want to take a similar approach

dabbott:
that's a good idea, and the only reason I didn't was cause the Arg API makes it kind of hassle 😢, but we can do it another time

Issue #372, opened 2019-04-10
File format spec: Text-transform in Text Style Definitions


ChristophLabacher:
Hi, not sure if this is the right place for this – if it isn’t please point me to the right one 😊

I started to work with the text style format defined by Lona in a project of mine and started to wonder, if the text-transform property should be an optional attribute of the text style spec as well. I have not made up my mind yet, but would love the see discussion of this.

My thinking is the following: While it is not a purely visual property, often times in modern UI design all-caps is used as a form of emphasis (e.g. for text labels), much like you would use the bold cut of a font. The "text-transform" can therefore very well be part of a specific type style.

I was wondering what other people think on this matter?

Best,
Christoph

dabbott:
Hi Christoph, you're in the right place!

I've run into a few cases where text-transform would be useful also, and I think it could be part of the spec. In design systems, it's pretty common to have text styles which should _only_ be used in uppercase, so I think it's reasonable for tools to enforce casing (optionally, for those who want it).

In terms of code generation:
- it's easy for web and RN, since it's built-in
- for iOS/macOS, we can add it to our `TextStyle` utility. we can't fully enforce it if people use the `attributeDictionary` API, since `NSAttributedString` doesn't support text transformations, but Lona doesn't use this API in the generated code at all, so I think that's fine https://github.com/airbnb/Lona/blob/4aa870a0502aa246c04ecea8cf09fed0c8d4e694/studio/LonaStudio/Generated/TextStyle.swift#L92
- for Android, we only have limited support for text styles at the moment, but it looks like there's a `textAllCaps` attribute for this, so we should be OK https://developer.android.com/reference/android/widget/TextView.html#attr_android:textAllCaps

I'd also be curious to hear more about what you're doing if you have a link to the project or want to chat sometime

ChristophLabacher:
I am experienced with web development and have a little experience writing Swift. What would be the best way for me to contribute in order to move this issue along?

dabbott:
There are 3 places we'll want to update:
- The spec doc
- Lona Studio
- The compiler

We can update each of these independently or all together, either way is fine. If you're most interested in specific part(s), you're welcome to just do those. I'm also happy to help or find somebody else to help (there's a person looking for a beginner-friendly compiler task, this might be one). What do you think?

Anyway, here's roughly what needs to be done for each:

### The spec

We should add an optional `textTransform` String to the Text Style type.

[...]

PR #389, opened 2019-05-28
#372 – Add TextStyles Property textTransform (WIP)

ChristophLabacher:
In this pull request I want to add support for the textTransform property for the TextStyles definition, as discussed in #372.

## To Do

- [ ] Update Docs
- [ ] Update Compiler
  - [ ] Implement TextTransform for JavaScript
  - [ ] Implement TextTransform for Swift (iOS)
  - [ ] Implement TextTransform for Swift (macOS)
  - [ ] Implement TextTransform for Android (not necessarily part of this pull request)
- [ ] Update Lona Studio

cc: @dabbott


ChristophLabacher:
@dabbott I started working on implementing this feature request. Since I am new to this project I thought it best to create a pull request to ask my questions here, even while it is still WIP.

I am not sure if this is the correct way of doing what you suggested:
https://github.com/ChristophLabacher/Lona/blob/add-textstyles-texttransform/compiler/core/src/static/swift/TextStyle.uikit.5.swift#L118-L160

This seems a bit hacky, but seemed the best option to me:
https://github.com/ChristophLabacher/Lona/commit/0ebc51d72a1598f0195d396500878dddc95466ea#diff-938a6276cbf8fcfa9a884e1917c8f74cR155

Implementing the transform for a specific range will probably be a bit more involved, so I wanted to check whether I am on the right path first.

dabbott:
This looks right to me. After looking at what you have so far, I'd suggest we make the `textTransform` member variable non-optional here: https://github.com/ChristophLabacher/Lona/blob/0ebc51d72a1598f0195d396500878dddc95466ea/compiler/core/src/static/swift/TextStyle.uikit.5.swift#L15
And the initializer will also use a non-optional type with a default of `.none`. That should simplify a bit of the logic later.

I'm thinking we should get rid of `capitalize`, since I really can't think of any use cases. The algorithm doesn't capitalize things well enough to be used for titles. But I don't feel very strongly about it either way. I just tried it in a playground and it looks like we should be able to call `capitalize` directly on a Swift String if we want it?
Screen Shot 2019-05-28 at 9 06 50 AM

If they're going to share some functionality, we could have `public func apply(to attributedString: NSAttributedString) -> NSAttributedString` call the version that uses range, `func apply(to attributedString: NSMutableAttributedString, at range: NSRange)` and only have to implement text transform once for both.

And/or it might come out slightly simpler looking if we have a separate function for applying the transformation to a mutable string:

```swift
private func apply(textTransform: TextTransform, to attributedString: NSMutableAttributedString, in range: NSRange) {
    switch textTransform {
    case .none:
        break
    case .uppercase, .lowercase, .capitalize:
        let substring = attributedString.mutableString.substring(with: range)
        let transformedSubstring = apply(textTransform: textTransform, to: substring)
        attributedString.mutableString.replaceCharacters(in: range, with: transformedSubstring)
    }
}
```

(no need to do extra work in the `none` case, which is the common case)

Minor style thing: I'd also try not to use `default` cases in switches, unless it causes awkward Swift 4 vs 5 issues. It'll be slightly easier to see we handle each case correctly if we handle each case specifically. There are some other misc little things that you'll run into when you try to compile it, but nothing important

ChristophLabacher:
Thanks for your feedback!

I feel like `capitalized` is a valid use case, but I have to think about it for a bit …

I tried to leave the function to apply text transform simple and put the substring functions into `apply(to attributedString: NSMutableAttributedString, at range: NSRange)`, since its the only place where we need it:

https://github.com/airbnb/Lona/pull/389/commits/a3c5674d0a749d5909c8b9a04234158885f25539

ChristophLabacher:
@dabbott I now got text transforms to display correctly in Lona Studio.

I defined the textTransform property as a String in the initializer arguments, and then turn it into an enumeration value in the initializer function.

https://github.com/airbnb/Lona/blob/27b5e84222fe0e72926a9fc39845e9c50475ae39/compiler/core/src/static/swift/TextStyle.appkit.swift#L35

It would probably be nicer to use the enumeration type as an argument, but then the swift compiler would have to transform the string into an enumeration (probably [here](https://github.com/airbnb/Lona/blob/27b5e84222fe0e72926a9fc39845e9c50475ae39/compiler/core/src/swift/swiftTextStyle.re#L82-L88)) and I don’t know how/if that’s possible.

dabbott:
Nice! This is great progress 😄

Editing the component file directly is probably the way to go for now. I'm working on some big updates to the input fields, but on master they're sort of a work in progress, and some types/values aren't selectable anymore.

In order to generate the text transform as an enum, I think we can do something similar to how font weight works: https://github.com/airbnb/Lona/blob/27b5e84222fe0e72926a9fc39845e9c50475ae39/compiler/core/src/swift/swiftTextStyle.re#L18-L28
https://github.com/airbnb/Lona/blob/27b5e84222fe0e72926a9fc39845e9c50475ae39/compiler/core/src/swift/swiftTextStyle.re#L49-L60

ChristophLabacher:
@dabbott Cool, I’ll look into that!

Can you tell me which View is used for displaying the TextStyles in the (new?) TextStyle picker? I can’t figure out whether it is `TextStyleRowView.swift`, `TextStylePickerView.swift`, `PickerListView.swift`, `TextStyleBrowser.swift` or something else – no matter which one I edit, the changes never appear in the picker. The text transform is displayed correctly in the text styles overview, but not in the picker, and I am trying to figure out why …

dabbott:
Ah, I forgot about that, it's part of a different package pulled in through Carthage (https://github.com/dabbott/Logic/blob/master/LogicDesigner/workspace/components/TextStyleRow.component). That component is generated with Lona and should update automatically once I re-generate the `TextStyle` class with these changes. I can do that once the rest of this is ready.

ChristophLabacher:
@dabbott Sorry for letting this slide, busy weeks …

So from what I can tell JavaScript, iOS and macOS should all function and Lona Studio should be fully updated one you regenerate the TextStyles for the Logic repo.

Please let me know if anything else needs to be done for this Pull Request.

PR #390, opened 2019-06-06
Support size tokens

dabbott:
## What

This adds experimental support for size tokens, and more generally arbitrary constants. I don't consider this "ready for use", so I'm not going to add docs yet.

Screen Shot 2019-06-05 at 8 07 53 PM


ryngonzalez:
YES!!!!

PR #410, opened 2019-10-31
updated readme to reflect compiler settings

grinder81:
## What

Updated Readme to show custom compiler settings.  

## Why

Current `.app` doesn't contain `lonac` by default and custom path doesn't have `ENV` path as well. Explained how to set it.

## Major Changes

* Updated Studio Readme and added a subsection

cc: @dabbott, @ryngonzalez


grinder81:
Looks great! We can merge it now.

PR #412, opened 2019-11-05
handle new md tokens in sketch library generation

mathieudutour:
Update the sketch library generation to handle the new tokens specified in markdown

### DIFF
--- a/compiler/sketch-library/lib/index.js
+++ b/compiler/sketch-library/lib/index.js
@@ -127,6 +127,9 @@ module.exports = function generateSketchLibrary(
         },
         tokens: tokens.files.reduce(
           (prev, file) => {
+            if (file.contents.type !== 'flatTokens') {
+              return prev
+            }
             file.contents.value.forEach(token => {
               if (token.value.type === 'color') {
                 prev.colors.push(token)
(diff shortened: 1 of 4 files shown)


compiler/sketch-library/lib/index.js
```diff
@@ -86,30 +86,64 @@ module.exports = function generateSketchLibrary(
...
-          workspace: compilerConfig.paths.workspace,
-          colors: compilerConfig.paths.colors,
-          textStyles: compilerConfig.paths.textStyles,
-          components: compilerConfig.paths.components,
+          workspace,
+          components: config.paths.components,
         },
+        tokens: tokens.files.reduce(
+          (prev, file) => {
+            file.contents.value.forEach(token => {
```
dabbott: It might be worth checking that `file.contents.type` is `"flatTokens"` to make this more resilient in case we end up including other kinds of converted content (SVGs or something?) in the flattened JSON

https://github.com/Lona/flatjson/blob/dad18051719477e80254d0044ccf2f5e45eb8410/flat-token-types.ts#L42

mathieudutour: Good idea 👍

PR #442, opened 2019-11-08
Use new tokens .md format

dabbott:
Token files are now stored in markdown files.

This is still experimental, but the PR is begin to get way too big, so I'm going to merge everything now.

This also adds the compiler `flatten` command and TypeScript type generation.


compiler/core/src/core/config.re
```diff
@@ -270,6 +285,7 @@ let load =
       ++ "'. A workspace must contain a `lona.json` file.",
     )
   | Some(workspacePath) =>
+    Log.warn("Running compiler from: " ++ dirname);
```
mathieudutour: shall this be using some kind of debug helper instead of always printing it? Something like https://github.com/visionmedia/debug#readme (not sure what's there for Reason)

dabbott: Yeah, that's a good idea

PR #444, opened 2019-11-09
Add the `documentation` command to the compiler

dabbott:
Run `lonac documentation --workspace [path]` to generate md pages to display on a docs site

This also updates the reason type generation and XML ast/print

### DIFF
--- a/compiler/core/src/main.re
+++ b/compiler/core/src/main.re
@@ -1,4 +1,6 @@
 open Node;
+open Monad;
+open Operators;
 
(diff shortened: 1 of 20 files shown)


compiler/core/src/main.re
```diff
@@ -1,4 +1,6 @@
 open Node;
+open Monad;
```
mathieudutour: ooh having fun with monad? 😄

dabbott: yeah 😂 this is a total misnomer though, since it only works with Reason's built-in optional type for now

PR #446, opened 2019-11-17
Fix #358: Allow to delete workspace file using "delete" key 

### LINKED ISSUE #358: [Lona Studio] Delete files by pressing "Delete" button
## Summary of Idea

At the moment, you can delete files in the file navigator only by using the context menu. It would be really nice to be able to delete files just by pressing "delete" button.

## Expected Behavior
A user should select a file and press "delete" button on the keyboard.

## Current Behavior
A user has to select file, right click it, and select "Delete" option.
Screen Shot 2019-03-27 at 6 56 39 PM

```

phimage:
## What

Delete files by pressing "Delete" button #358 on file tree
The event is triggered by FileTree https://github.com/dabbott/FileTree/blob/master/FileTree/FileTree.swift#L307 , so the key is backspace

This show an alert, same factorised code as delete menu item

## Why

I did this change in order to allow power user to use keyboard

cc: @dabbott, @klymenkoo

### DIFF
--- a/studio/LonaStudio/Workspace/FileNavigator.swift
+++ b/studio/LonaStudio/Workspace/FileNavigator.swift
@@ -132,6 +132,7 @@ class FileNavigator: NSBox {
         fileTree.filterFiles = { path in
             return !(path.hasPrefix(".") || URL(fileURLWithPath: path).deletingPathExtension().path.hasSuffix("~"))
         }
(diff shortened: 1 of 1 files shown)


dabbott:
Perfect! Thanks for the contribution 😀

PR #447, opened 2019-11-19
update react-sketchapp

mathieudutour:
## What

- [x] update react-sketchapp to v3
- [x] move `sketch-library` to TS to leverage the new types from react-sketchapp and sketch-file


dabbott:
:+1: Nice!

PR #453, opened 2020-01-08
move serializer to typescript

mathieudutour:
While going through the `LogicFile.convert(markdownData, kind: .document, to: .json, from: .source)` bug, I ended up in `compiler/serialization` and couldn't quite make sense of it.

So I grouped things by logical dependency and now it seems more approachable (IMO).

Let me know if it makes sense.

I always moved it to typescript to understand what was going through the functions and there is one key type missing: the result of `logicSwiftParser.parse`. I guess I could create it from the peg grammar but it's been a while since I touched any so I don't really remember how it works.

I also made sure that all the markdown nodes were printed correctly (in `compiler/serialization/src/mdast/toMarkdown.ts`)

cc: @dabbott


compiler/serialization/bin.js
```diff
@@ -9,35 +9,36 @@ const {
   convertLogic,
   convertDocument,
   extractProgram,
-} = require('./lib/index')
```
mathieudutour: pretty sure this was a bug as the `lib` folder isn't included in the npm package

mathieudutour:
I made another change: instead of creating a new node type for markdown, it set a flag on `page: true` on the link. That way we keep a proper mdx AST. (the intermediate AST still have a `page` node)

compiler/serialization/bin.js
dabbott: oh yeah, that seems likely

PR #457, opened 2020-01-09
Sync directory structure with markdown page links

dabbott:
## What

Here are the operations we'll attempt to sync:
- Creating a child page (Child.md) block within a page (Page.md)
  - [X] Convert Page.md to a directory containing a README.md, insert child page block in README.md, create Child.md page
  - [x] Update the parent's URL to point to Page instead of Page.md
- Deleting a child page block
  - [X] Delete Child.md, remove the child page block from README.md
  - [x] If Page/README.md contains no children: If it's safe to delete Page (a directory containing only a README.md and/or .DS_Store), convert Page/README.md to Page.md
- [x] Update selection state in file navigator when opening files
- [x] Use page icon for a directory with a README.md


mathieudutour:
> Update the parent's URL to point to `Page` instead of `Page.md`

Couldn't it always point to `Page`?

mathieudutour:
A bit unrelated but do you think it'd be possible to select the file in the sidebar when navigating to it? It feels weird to change files but still have the parent selected

dabbott:
> Couldn't it always point to Page?

That would break linking between pages on Github though

> Do you think it'd be possible to select the file in the sidebar when navigating to it? It feels weird to change files but still have the parent selected

:+1: definitely. I forgot to add that to the list

> A folder with a readme probably shouldn't have a folder icon in the sidebar. It's still a page

:+1: Thinking out loud, if a folder has other stuff in it, like image assets, maybe we'd want to show an asset "subicon" on the page icon to indicate that it's a page+assets

mathieudutour:
> That would break linking between pages on Github though

Ah right good point

> maybe we'd want to show an asset "subicon" on the page icon to indicate that it's a page+assets

Is the arrow indicating that there are nested stuff inside not enough?

dabbott:
> Is the arrow indicating that there are nested stuff inside not enough?

Maybe that's enough... after thinking about it more, I guess the subicon does't really solve anything. 

I feel like there's a case here where you'd want to put the same thing on multiple pages, e.g. image assets, and so it feels weird that one page would "own" them. I guess if a folder doesn't have a readme then it's a regular folder? but, if it does have a readme, it seems like things could feel hidden. or maybe it'll be fine, idk

mathieudutour:
So that a good question.

- Do we want to hide folders that do not have a readme?
  If no then they would have a normal folder icon
- Do we want to hide files that aren't linked from the readme?
  If no, then they would indeed be a bit hidden from the readme. But maybe we can have a section in the readme preview which would act as a "gallery" of files that aren't linked from the rest of the readme (kind of what we had for the root of the workspace before)

dabbott:
@mathieudutour I think this is in a decent spot if you want to take a look. There are a _lot_ more cases to consider and bugs to work out, but I think this works well enough to start playing with.

mathieudutour:
I'll play with it tomorrow 👍

mathieudutour:
It's pretty good.

- how do you rename a page? Should changing the h1 in the subpage change the name of the page? Instinctively, I'd single click to rename (eg. select the node), and double click to navigate to the subpage.
- I'm missing a swipe with 2 fingers to go back (like on the web)
- Small detail but it'd be nice to have a breadcrumb at the top instead of just the current page (although I think it's the default macOS behaviour? I'm not sure if there is an API for it)
- when you hover on the page block, there are a "three dots" icon that appear but doesn't do anything
- Shall it be possible to drag and drop a subpage to another part of the file tree?
- if I manually delete a link but keep the subpage, and then try to add a subpage with the same name, it shows 2 errors ("can't create page" + "can't save file") but I think it should just add the link without complaining

dabbott:
> how do you rename a page? Should changing the h1 in the subpage change the name of the page? Instinctively, I'd single click to rename (eg. select the node), and double click to navigate to the subpage. Or maybe the opposite 🤔

I like the idea of having the h1 change the name of the page, although, I'd probably need to make editing it a bit special/structured (e.g. not rename on every keystroke, not allow empty string, not allow bolding/italic) so I think I'm gonna come back to that another time.

Single clicking these page links to select?

Screen Shot 2020-01-24 at 4 10 19 PM

I'm not really sure what the right behavior is... for now I'm copying notion's behavior with single click to navigate, and I'm gonna put rename in the "..." menu. it might be confusing if it did something different than notion, since most other things are the same.

> I'm missing a swipe with 2 fingers to go back (like on the web)

Oh yeah, that's a good idea. history/back in general feels important

> Small detail but it'd be nice to have a breadcrumb at the top instead of just the current page (although I think it's the default macOS behaviour? I'm not sure if there is an API for it)

A breadcrumb would be super helpful. We're not really using the mac APIs for the title (it's actually a Lona component) so can definitely do a breadcrumb

> when you hover on the page block, there are a "three dots" icon that appear but doesn't do anything

I'm gonna put rename in there I think. Delete too, for consistency

> Shall it be possible to drag and drop a subpage to another part of the file tree?

Yeah, I disabled it at the moment, but definitely

> if I manually delete a link but keep the subpage, and then try to add a subpage with the same name, it shows 2 errors ("can't create page" + "can't save file") but I think it should just add the link without complaining

I agree. I think there are some other cases like this where manually deleting stuff can leave you in a weird state, and we can handle it smarter than it is right now

mathieudutour:
> not rename on every keystroke, not allow empty string, not allow bolding/italic

Right, renaming on blur if the string is not empty should be fine

> for now I'm copying notion's behavior with single click to navigate, and I'm gonna put rename in the "..." menu

Ah ok, somehow it felt different from Notion, maybe because it’s so fast haha

PR #459, opened 2020-01-11
Compiler supports converting md files in workspace to js

dabbott:
## What

- The `workspace` command will now convert md files to js
- Added file snapshots for JS/Swift/Documentation generated from tokens
- Removed the "// Generated by ..." banner on snapshot files for less noisy diffs (in the future)

Note: The generated JS probably isn't exactly what we want, but it should at least be valid JS

@mathieudutour


mathieudutour:
One thing to note is that we should either generate `.mjs` files or compile the js files. You can't put `export let x =''` in a node package.

dabbott:
Ah, good point. Hmm... I guess I'll use .mjs for now, and if it becomes an issue for people (or us) at some point, we can do something else

mathieudutour:
👍 We will still need compile jsx but we can do that "manually", no need for babel.

PR #462, opened 2020-01-15
start to refactor compiler

mathieudutour:
## What

refactor compiler

cc: @dabbott


mathieudutour:
moved to https://github.com/Lona/compiler

PR #474, opened 2020-03-17
Automatically configure SSH key for user

dabbott:
This needs some testing still, but the idea is:

Can we make a fully automated flow, so that once somebody has a GitHub account, we can set up push access? (everything is opt-in)

This means, if a push fails due to invalid credentials:
- we prompt them to ask if we should set up SSH access
- if they don't have an .ssh key, we create one
- we upload their public key to their github account (which requires `write:public_key`)
- we verify the identity of github.com using a compiled-in SSH fingerprint

I couldn't find a GraphQL mutation for adding a public_key, so I use the API v3 here... but maybe I just missed it?

I thought it was going to be hacky but it's actually not too bad IMO. There's still a bug after github repo creation where it doesn't move to the next part of the flow automatically

### DIFF
--- a/studio/LonaStudio/Utils/RESTClient.swift
+++ b/studio/LonaStudio/Utils/RESTClient.swift
@@ -59,6 +59,16 @@ extension RESTClient {
         client.configureRequest = { request in
             var request = request
 
+            let group = DispatchGroup()
+            group.enter()
+
+            Account.shared.me().finalResult({_ in
+                group.leave()
+            })
+
+            // wait for the Lona account to be fetched
+            group.wait()
+
             if let githubToken = Account.shared.cachedMe?.githubAccessToken {
                 request.addValue("application/vnd.github.v3+json", forHTTPHeaderField: "Accept")
                 request.addValue("Bearer \(githubToken)", forHTTPHeaderField: "Authorization")
--- a/studio/LonaStudio/Utils/RESTClient.swift
+++ b/studio/LonaStudio/Utils/RESTClient.swift
@@ -1,0 +1,95 @@
+//
+//  RESTClient.swift
+//  LonaStudio
+//
+//  Created by Devin Abbott on 3/16/20.
+//  Copyright © 2020 Devin Abbott. All rights reserved.
+//
+
+import Foundation
+
+class RESTClient {
+    init(baseURL: URL) {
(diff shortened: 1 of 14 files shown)


studio/LonaStudio/VersionControl/Git.swift
```diff
@@ -79,81 +101,97 @@ public enum Git {
...
         return run(arguments: ["remote", "add", name, url.absoluteString])
     }
 
-    public func push(repository: String, refspec: String) -> Result<String, NSError> {
-        return run(arguments: ["push", repository, refspec])
+    public func push(repository: String, refspec: String) -> Result<String, GitError> {
+        return run(arguments: ["push", repository, refspec]).mapError({ error in
+            if error.code == 128,
+                let errorString = error.errorOutput.utf8String(),
+                (errorString.contains("git@github.com: Permission denied (publickey)") || errorString.contains("Host key verification failed")) {
```
mathieudutour: in dugite (what GitHub is using), they just check that the code is !== 0 and that's their regex: https://github.com/desktop/dugite/blob/e447c5ef95b2cfd13e98a4d85f89e7906f82dbdc/lib/errors.ts#L60-L63

mathieudutour:
That's really cool 💯

studio/LonaStudio/Utils/RESTClient.swift
```diff
@@ -0,0 +1,95 @@
...
+}
+
+extension RESTClient {
+    static var githubV3: RESTClient = {
+        let client = RESTClient(baseURL: URL(string: "https://api.github.com")!)
+
+        client.configureRequest = { request in
+            var request = request
+
+            if let githubToken = Account.shared.cachedMe?.githubAccessToken {
```
mathieudutour:
```suggestion
            let group = DispatchGroup()
            group.enter()

            Account.shared.me().finalResult({_ in
                group.leave()
            })

            // wait for the Lona account to be fetched
            group.wait()

            if let githubToken = Account.shared.cachedMe?.githubAccessToken {
```

studio/LonaStudio/VersionControl/Git.swift
dabbott: :+1: copying theirs

PR #478, opened 2020-03-22
update graphQL schema

mathieudutour:
## What

- [x] add `name`, `description`, and `openGraphImageUrl` to a lona repo. They might be `nil` (for all existing repos in the db)


dabbott:
I restarted the build job, but since the PR build passed I don't see any problem with merging as-is

PR #479, opened 2020-03-25
Perf low hanging fruits

mathieudutour:
## What

- [x] only invalidate the cache when there is an actual change
- [x] only invalidate the cache when the programs are non-equivalent


studio/LonaStudio/Documents/LogicDocument.swift
```diff
@@ -73,10 +73,6 @@ class LogicDocument: NSDocument {
 
     override func read(from data: Data, ofType typeName: String) throws {
         content = try LogicDocument.read(from: data)
-
-        if let url = fileURL {
-            LogicModule.invalidateCaches(url: url, newValue: program)
```
mathieudutour: this is already done when setting `content`

dabbott: Good catch

studio/LonaStudio/Documents/LogicDocument.swift
```diff
@@ -92,8 +88,15 @@ class LogicDocument: NSDocument {
...
     override func save(to url: URL, ofType typeName: String, for saveOperation: NSDocument.SaveOperationType, completionHandler: @escaping (Error?) -> Void) {
+        let dataOnDisk = try? Data(contentsOf: url)
+
         super.save(to: url, ofType: typeName, for: saveOperation, completionHandler: completionHandler)
 
-        LogicModule.invalidateCaches(url: url, newValue: program)
+        let newData = try? self.data(ofType: typeName)
+
+        // only invalidate if what was on the disk is different from what we saved
+        if dataOnDisk != newData {
```
dabbott: Now I'm curious if we can move the `super.save(to: url, ofType: typeName, for: saveOperation, completionHandler: completionHandler)` to within this `if`, avoiding an unnecessary write. doesn't need to be part of this, I'm not sure if it'd caused problems or not

PR #483, opened 2020-04-06
Component editor UI updates

dabbott:
This improves the component editor UI by moving the element editor into the left sidebar, and ensuring colors look good in both light and dark mode


studio/LonaStudio/Workspace/LayerList.swift
```diff
@@ -64,12 +64,13 @@ public class LayerList: NSBox {
         boxType = .custom
         borderType = .noBorder
         contentViewMargins = .zero
-        fillColor = .red
 
+        fillColor = Colors.vibrantWell
+
```
lonabot, line 69: - :warning: Lines should not have trailing whitespace. (`trailing_whitespace`)

PR #485, opened 2020-04-07
Misc publishing and onboarding UI improvements

dabbott:
- Improved form submit state
- Improve FileNavigator initial state
- Reopen welcome window when the workspace window is closed
- Bump logic and filetree


mathieudutour:
> Reopen welcome window when the workspace window is closed

When I did the change, I intentionally did not do that. I checked a few apps that have a welcome screen and they don't reopen it when closing a document

dabbott:
Interesting... it feels better like this to me though, and I can't really think of any downside to it

PR #486, opened 2020-04-08
Persist window frame

dabbott:
The workspace window will maintain the same size and position across launches


mathieudutour:
AppKit should already handle it if you set the [`frameAutosaveName `](https://developer.apple.com/documentation/appkit/nswindow/1419362-frameautosavename) on the NSWindow

dabbott:
I wasn't able to get that working last time I tried (in SourceNote), so I just copied what I did there. It might work fine though. I remember following https://stackoverflow.com/questions/2734888/how-can-i-easily-save-the-window-size-and-position-state-using-obj-c/44721893#44721893 at the time

PR #487, opened 2020-04-08
add docs url

mathieudutour:
## What

- [x] update the `getMe` query to get the docs URL
- [x] add a `getRepo` to check if a repo is already activated


dabbott:
Is this ready to go, or does it need updates?

mathieudutour:
it is now!

mathieudutour:
They do, but it doesn't change anything just yet

PR #488, opened 2020-04-09
Add Page outline view

dabbott:
This adds a page outline view to the left sidebar, which helps for navigating large pages.

Screen Shot 2020-04-08 at 8 20 22 PM


mathieudutour:
I don't think it needs the `H1 -` prefix. 

But I'm wondering how it will scale with workspaces that have both a lot of pages and large pages. Wouldn't it make 2 small scrollviews, none really useful to have a good overview?

How about leveraging the horizontal space (which we have more of than vertical space) to show a minimap on the right?

dabbott:
Yeah, it's definitely not perfect. maybe I'll sit on it for a bit and see if I can come up with something better.

I like the idea of a minimap. it might be a little more effort to make it fast, but maybe not.

I think as long as both the scrollviews are collapsible it'd be OK real-estate-wise - e.g. VSCode. 
the `H1 -` helps disambiguate in cases like this, e.g. the initial state of a blank doc with just a title, where it's not clear how the title differs from the file name. maybe something else would be better.

Screen Shot 2020-04-09 at 12 48 14 PM

I _do_ like having an outline for long-form content, which is a bit different than a (traditional) minimap. (Xcode's minimap-outline hybrid is clever, but I never remember/use the outline feature)

I also like that it mirrors the elements outline of a component file

dabbott:
Ok! here's what I landed on:

Screen Shot 2020-04-11 at 5 43 40 PM

I'm liking the sections numbers as part of the outline a lot - maybe need an option to render them in the actual content at some point. the bottom scrollview doesn't collapse yet, but I think I can deal with that once we have bigger workspaces.

mathieudutour:
Looking good! 

Yeah my comment about `H1` was that users might not know what the `H` is for (IMO it's kind of a dev terminology). Happy with just `1` 👍

dabbott:
:+1: yeah, and even for those who do know what H1 is, it didn't add much. This is better

